PR #2, opened 2015-07-09
atm

aag6z:
comfort: 5
completeness: 3
i didn't get to all of the features; was not feeling well yesterday and decided to call it an early night

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,63 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingBalance_div = document.querySelector("div#checking_balance");
+var savingsBalance_div = document.querySelector("div#savings_balance");
+checkingBalance = 0;
+savingsBalance = 0;
+console.log(checkingBalance);
 
 var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+  checkingBalance += parseInt(amount);
+  checkingBalance_div.innerHTML = "$" + checkingBalance;
+  if(checkingBalance >= 0){
+    console.log("hi");
+    $("div#checking_balance").removeClass("zero");
+  }
+
 
 }
 
 function checkingWithdrawal(amount){
+  checkingBalance -= parseInt(amount);
+  checkingBalance_div.innerHTML = "$" + checkingBalance;
+  if(checkingBalance <= 0){
+    console.log("hi");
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,63 @@
...
 
 var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+  checkingBalance += parseInt(amount);
+  checkingBalance_div.innerHTML = "$" + checkingBalance;
+  if(checkingBalance >= 0){
+    console.log("hi");
+    $("div#checking_balance").removeClass("zero");
+  }
+
```
andrewsunglaekim, line 14: I love the use of console.log() throughout your program. One thing I would comment on, however, is to provide semantics for your logs. They should be stating what is happening when that part of the code is executing. As it stands in this current app what you did is fine. But if you were to debug a bigger problem on a grander scale you'll be using way more loggings to figure something out. That's where semantic logs will really help out

atm.js
```diff
@@ -1,21 +1,63 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingBalance_div = document.querySelector("div#checking_balance");
+var savingsBalance_div = document.querySelector("div#savings_balance");
+checkingBalance = 0;
+savingsBalance = 0;
+console.log(checkingBalance);
```
andrewsunglaekim, line 5: I think instantiating two values of 0 is fine for this exercise. But eventually will figure out how to persist data in our applications and the balance wont always start out this way. I think better might be to access the balance through the DOM itself.

atm.js
aag6z, line 14: Thank you. That actually makes a lot of sense. I"ll try doing that for the project I'm working on right now. I think it'll be better than "hi" lol

atm.js
```diff
@@ -1,21 +1,63 @@
...
+  checkingWithdrawal($("#checking_amount").val())
+});
+
+$("#savings_withdraw").on("click", function(){
+  savingsWithdrawal($("#savings_amount").val())
+});
+
+$("#savings_deposit").on("click", function(){
+  savingsDeposit($("#savings_amount").val())
+});
```
andrewsunglaekim, line 63: love this, nice and clean. You could also just pass the callback and its argument as the function instead of using an anonymous function, eliminating 8 lines of code.

andrewsunglaekim:
Good job on the HW! Nice and clean, let me know if you have any questions.

PR #3, opened 2015-07-09
Tom beach solution atm

tomBeach:
comfortLevel: 4
completeness: 5

added comments, removed console.logs, fixed display formatting problem (submitted 7pm thurs)

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,175 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
+// ======= init
+var checkingTotal = 0;
+
+// ======= balance fields
+var Cbalance = parseInt($('.balance').eq(0).html().substring(1));
+var Sbalance = parseInt($('.balance').eq(1).html().substring(1));
+
+// ======= buttons
+var Cdeposit = $('.account input').eq(1).on('click', function(event) {
+  make_transaction('checking', 'deposit');
+});
+
+var Sdeposit = $('.account input').eq(4).on('click', function(event) {
+  make_transaction('savings', 'deposit')
+});
+
+var Cwithdraw = $('.account input').eq(2).on('click', function(event){
+  make_transaction('checking', 'withdraw');
+});
+
+var Swithdraw = $('.account input').eq(2).on('click', function(event){
+  make_transaction('savings', 'withdraw');
+});
+
+function make_transaction(whichAccount, dep_Or_wdrw) {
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,99 @@
...
+	make_transaction('savings', 'deposit')
+});
 
-}
+var Cwithdraw = $('.account input').eq(2).on('click', function(event){
+	make_transaction('checking', 'withdraw');
+});
+
+var Swithdraw = $('.account input').eq(5).on('click', function(event){
+	make_transaction('savings', 'withdraw');
```
adambray, line 22: I really like how for each event listener, you just call one function, this really keeps your UI code nice and short/tidy.

atm.js
```diff
@@ -1,21 +1,99 @@
...
+});
+
+
+// ======= make_transaction =======	
+function make_transaction(whichAccount, dep_Or_wdrw) {
+
+	// ======= input fields ======= 
+	var CdepositAmount = $('.account input').eq(0).val();
+	var SdepositAmount = $('.account input').eq(3).val();
```
adambray, line 31: Unless I'm looping over items, I generally prefer to use classes or id selectors instead of `eq`, as it makes the code easier to read (and less fragile if the HTML changes).

atm.js
```diff
@@ -1,21 +1,99 @@
...
+
+	// ======= validation (text entry) ======= 
+	if ((isNaN(CdepositAmount) && CdepositAmount != '') || (isNaN(SdepositAmount) && SdepositAmount != '')) {
+		$('.balance').eq(0).addClass("message");
+		$('.balance').eq(0).html('numbers only please');
+	} else {
```
adambray, line 37: Anytime I have some code that does a single job, I like to break it out into a separate function, i.e. here I might move this code into a function `validateNumerical` and then call that function here instead. Makes the code more readable, maintainable, and flexible.

atm.js
```diff
@@ -1,21 +1,99 @@
...
+						$('.balance').eq(0).html('cancelled: negative account balance');
+					}
+
+				// ======= enough to cover =======		
+				} else {
+					$('.balance').eq(0).removeClass("message");
+					$('.balance').eq(0).html('$' + Cbalance);
+				}
+
+			// ======= savings =======
```
adambray, line 81: I really appreciate how well commented and formatted your code is, but the fact that there are many nested `ifs` is a signal that this code might benefit from a refactoring.

I think the long-term goal would be to use an object to represent an account (with properties like `balance`, and methods like makeDeposit(amount) and makeWithdrawl(amount)`. You would have two objects, one for each account, and reference those objects and their methods in the click handlers.

I'm happy to pair with you on how we might do this during our next one-one one. You could also reference the object-based solution to Timer JS we did during the review session.

adambray:
Good job! See my line comments above, and let me know if you have any questions! :+1:

PR #5, opened 2015-07-09
Jared solution

jaredarkin:
comfort_level: 5
completeness: 5

i think this can be done using an object-oriented approach but i'm having trouble getting started. i'll probably start another branch to work on it.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,10 +1,17 @@
-/*var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingBalance = document.querySelector("div.checking_balance").innerHTML;
+/*checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );*/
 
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,59 @@
...
+  var totalBalance = balance + savingsBalance;
+  if( newWithdrawal < balance ){
+    var newBalance = balance - newWithdrawal;
+    $("#checking_balance").text("$" + newBalance);
+  } else if ( newWithdrawal <= totalBalance){
+    $("#checking_balance").text("$0");
+    $("#checking_balance").addClass("zero");
+    var newBalance = totalBalance - newWithdrawal;
+    $("#savings_balance").text("$" + newBalance);
+  }
```
adambray, line 28: Probably want to handle the `else` case here by notifying that the withdrawal couldn't be made due to not enough in savings + checking!

adambray:
This all looks great! :100: 

I'm happy to pair with you on the OOP approach if you're interested.

PR #6, opened 2015-07-09
Mattficke solution

mattficke:
comfort: 5
completeness: 4 (it's not yet particularly object oriented, will refactor)

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -13,7 +13,7 @@ var checkingDeposit = function(amount){
 
 function checkingWithdrawal(amount){
   var amount = parseInt($("#checkingAmount").val());
-  if (amount) {
+  if (amount && amount <= checkingBalance) {
     checkingBalance = checkingBalance - amount;
   }
   $("#checkingBalance").html("$"+checkingBalance);
@@ -31,7 +31,7 @@ function savingsDeposit(amount){
 
 function savingsWithdrawal(amount){
   var amount = parseInt($("#savingsAmount").val());
-  if (amount) {
+  if (amount && amount <= savingsBalance) {
     savingsBalance = savingsBalance - amount;
   }
   $("#savingsBalance").html("$"+savingsBalance);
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,105 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingBalance = 0;
+var savingsBalance = 0;
```
andrewsunglaekim, line 3: I've seen this alot, and while its fine for this exercise. I would say that its not entirely modular. Imagine we have the ability to persist data and that we could save the state of the balances. If we instantiate the balances to 0 every time in our JS it would do just that despite what the balance is saved at.

Additionally, you would need to make a function `getBalance()`, because you would have to retrieve it for every event that happened in which you need it.

atm.js
```diff
@@ -1,21 +1,105 @@
...
+  else {
+    $("#checkingBalance").removeClass("zero");
+  }
+
+  if (savingsBalance === 0) {
+    $("#savingsBalance").addClass("zero");
+  }
+  else {
+    $("#savingsBalance").removeClass("zero");
+  }
```
andrewsunglaekim, line 77: using the falsey nature of 0 will eliminate some characters here  from lines 64 - 77 but you would have to switch the conditions IE:

```
if (checkingBalance){
// remove the class zero
}
```

atm.js
```diff
@@ -1,21 +1,105 @@
...
+
+$("#savingsDeposit").on("click", function() {
+  savingsDeposit();
+  balanceCheck();
+});
+
+$("#savingsWithdrawal").on("click", function() {
+  savingsWithdrawal();
+  balanceCheck();
+});
```
andrewsunglaekim, line 101: not sure if it's better to call balanceCheck inside of the withdrawal and deposit functionality. The more and more i think about it the more ambivalent about it i am.. hah

andrewsunglaekim:
Great job on the HW. I wouldn't be too caught up on an OO approach to this. We'll be getting into OOJS post week 7. Although we can take an object oriented approach to this solution. OO design makes alot more sense when were dealing with models/entities in a database. The solution for ATM takes a more OOP approach though. But good job either way :smile:

PR #7, opened 2015-07-09
ATM Homework

pdoulatshahi:
Comfort level: 5/5
Completeness: 5/5


amaseda:
Overall looks great. The ATM is totally functional! Next step here is to make your code DRY-er and get rid of repeated lines. If you were to take this to the extreme, you could create withdraw/deposit functions that work on any account, whether that's checking, savings or an additional account the user creates down the line.

PR #8, opened 2015-07-09
added features for all user stories

phantomhaircuts:
Comfort level: 4;
Completness: 5;

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,11 +1,105 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+//var checkingBalance = document.querySelector("div#checking_balance");
+//checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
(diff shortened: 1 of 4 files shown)


atm.js
```diff
@@ -1,11 +1,105 @@
...
+  console.log("zero")
+  $(".balance").eq(0).addClass("zero")
+}
+if (savingsBalance === 0){
+  console.log("zero")
+  $(".balance").eq(1).addClass("zero")
+}
+
+//DEPOSIT IN CHECKING
+$(":button").eq(0).on("click", function checkingDeposit(){
```
adambray, line 27: Probably better to use ids on the buttons to avoid using `.eq(0)` etc; this way, your code is more readable!

adambray:
Hey Erik!

I'd like to finish giving feedback on this assignment, but the indentation (esp. with the `if` statements) is a bit messed up, and it's hard to follow the logic. Would you mind fixing the indentation, and then add/commit/push your changes. That should auto update the PR, and I'll finish giving feedback asap!

-Adam

PR #10, opened 2015-07-10
ATM

kristingreenslit:
Comfort Level: 2
Completeness: 1

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,13 +1,24 @@
+//add event listener to listen for click on deposit button
+//when deposit button is clicked, change div with balance
 
-// $("input").val();
-$(".deposit_button").on("click",function(){
-  console.log("does it work?")
+ $(".Cdepbutton").on("click", function(){
+  var userEntry = $(".Ctext").val();
+  console.log(userEntry);
+  $('.balance').eq(0).val(userEntry);
 });
+  // var checkingDeposit = function(amount){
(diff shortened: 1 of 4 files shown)


atm.js
```diff
@@ -1,21 +1,19 @@
...
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingBalance = 0;
+var savingsBalance = 0;
 
-}
+$("#c-depbtn").on("click", function(){
+  var userInput = $("#c-text").val();
+  checkingBalance += userInput;
+  $("#c-balance").html(checkingBalance);
+});
```
andrewsunglaekim, line 11: i think you're pretty close here! Couple of notes. the `.val()` will always return a string. so if you were to add `"22" + 22` you would get `"2222"` That would make any bank very sad. I think what you want to do is parse it into an Integer first using `parseInt()`

atm.js
```diff
@@ -1,21 +1,19 @@
...
 
-function checkingWithdrawal(amount){
+$("#s-depbtn").on("click", function(){
+  var userInput = $("#s-text").val();
+  savingsBalance += userInput;
+  $("#s-balance").html(savingsBalance);
```
andrewsunglaekim, line 15: I think here you would want to use `-=` because we want to decrement

andrewsunglaekim:
Hey kristin i think this is alot better than what you had before. Additionally i'm really happy about your resubmissions of TTT and timer! Great work!

PR #11, opened 2015-07-10
Jeffs solution

jeffjones89:
Comfort: 4.5
Completeness: 4.5 

Committing this now, as checking/savings and class change when either account = $0 are working. Going to try and reconfigure my code to make overdraft protection more straightforward.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,10 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+cDeposit = $('#checking_deposit')
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(cDeposit).on('click', function (event){
 
-}
+    var currentBalance = $('.balance').eq(0);
+    depositInput = $('#checking_input');
+    newBalance = ('$' + (depositInput + currentBalance.html())
 
-function checkingWithdrawal(amount){
-
-}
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,64 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready (function(){
```
jshawl, line 1: the document.ready here works, but is unecessary. Because this file is loaded just before the closing body tag, the DOM will have already loaded

atm.js
```diff
@@ -1,21 +1,64 @@
...
+$('#checking_deposit').on('click', function() {
+    cInput =  $('#checking_input');
```
jshawl, line 3: remember to use `var` to prevent global variables

atm.js
```diff
@@ -1,21 +1,64 @@
...
+    currentSavings = sBalance.html();
+    $(sBalance).removeClass('zero')
+    sInput.val('')
+})
+
+$('#savings_withdraw').on('click', function() {
+    sInput =  $('#savings_input');
+    sBalance = $('.balance').eq(1)
+
+
```
jshawl, line 51: Your solution looks good! please remove empty lines/white space in your code

jshawl:
:+1: Nice job on this assignment! If you wanted to keep working on this, you might want to think about how you can organize your code so that every account doesn't need its own function. i.e. how could you allow users to add another checking account or savings account without having to hardcode more js click handlers

PR #12, opened 2015-07-10
atm Solution

ghost:
Comfort:4/5
Completeness: 5/5

JQuery made me understand Javascript more! I think I had problems with Javascript because I didn't break the solution down into micro steps. JQuery helped me understand it somehow (...weird). I think it's because it's a lot simpler than Javascript.


atm.js
```diff
@@ -1,21 +1,68 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready (function(){
```
jshawl, line 2: This is not wrong, but `document.ready` is unecessary here because your script tag is just before the closing body tag. the DOM will have loaded by the time your script is loaded

atm.js
```diff
@@ -1,21 +1,68 @@
...
+  //Checkings account
+    //when click on deposit,
+      //enter and store amount,
+      //show amount entered
+      //add from the total amount
```
jshawl, line 7: EXCELLENT pseudo code!!!

atm.js
```diff
@@ -1,21 +1,68 @@
...
+  //withdraw more money than exists in the account, ignore the transaction.
 
-function checkingWithdrawal(amount){
+var cTotal = 0;
+var sTotal = 0;
 
-}
+//Checking
+  $("#cDepositButton").on("click", function() {
+    var cDepositUserInput = parseFloat($("#cPlaceHolder").val());
```
jshawl, line 23: nice use of `parseFloat` here. the pennies matter too!!

atm.js
```diff
@@ -1,21 +1,68 @@
...
+  });
 
-function savingsDeposit(amount){
+  $("#cWithdrawButton").on("click", function() {
+    var cWithdrawUserInput = parseFloat($("#cPlaceHolder").val());
+    if (cTotal >= cWithdrawUserInput) {
+      cTotal -= cWithdrawUserInput;
+      $("#cBalance").text("$ " + cTotal);
+      //Make sure the balance in an account can't go negative. If a user tries to
+      //withdraw more money than exists in the account, ignore the transaction.
```
jshawl, line 34: awesome code comments! this makes it easy to read your code

atm.js
```diff
@@ -1,21 +1,68 @@
...
+    }else if (cTotal < cWithdrawUserInput) {
+      var answer = confirm("You don't have enought money. Would you like to withdraw from savings?");
+        if ("ok") {
```
jshawl, line 37: This will always return true. instead, `if ( answer == "ok")`

jshawl:
:+1: Glad to hear you like jQuery! I love how much you can do with so little code!

Nice job on this assignment. Can you think of a way to handle the transactions and update the html without separating the checking and savings functionality? i.e. if you wanted users to be able to have any number of accounts - what would the Javascript for that look like?

PR #13, opened 2015-07-10
ATM HW

Ryan-321:
comfort 3;

completeness 4;

I think I had it working better before I started to fix bugs and tried to clean up the code.  Still buggy.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,4 +1,10 @@
 
+var atm = {
+
+  
+
+}
+
 
 // Listeners
 
@@ -41,8 +47,10 @@ function WithdrawC() {
     $( "#checking .balance").addClass( "zero");
   }
 
-  if (sumChecking < 0 && sumSavings > 0) {
-    if (sumSavings < 0) {
+  if (sumChecking < 0 && sumSavings >= 0) {
+    if (sumSavings <= 0) {
+      $( "#savings .balance").addClass( "zero");
+      sumChecking = 0;
       return alert("You don't have enough money for this transaction.")
     }
     // Overdraft amount ( ex. -10 + 50 = 40)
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,86 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var atm = {
```
jshawl, line 2: Excellent attempt at implementing an object-oriented solution!

atm.js
```diff
@@ -1,21 +1,86 @@
...
+      $( "#savings input[value=Withdraw]" ).on("click" , atm.functions.WithdrawS );
 
-function checkingWithdrawal(amount){
+},
 
-}
+functions: {
+  DepositC: function() {
+            var depositChecking = $( "#checking input[type=text]" ).val();
+            atm.sumChecking += parseInt(depositChecking);
```
jshawl, line 17: if you use `this` instead of `atm` here, you don't have to worry about some other global `atm` variable that might affect your code.

atm.js
```diff
@@ -1,21 +1,86 @@
...
+      $( "#checking input[value=Withdraw]" ).on("click" , atm.functions.WithdrawC );
+      $( "#savings input[value=Deposit]" ).on( "click" , atm.functions.DepositS );
+      $( "#savings input[value=Withdraw]" ).on("click" , atm.functions.WithdrawS );
 
-function checkingWithdrawal(amount){
+},
 
-}
+functions: {
+  DepositC: function() {
```
jshawl, line 15: JavaScript uses Capital function names to indicate contructor functions. Prefer lower case unless creating the object with the `new` keyword.

atm.js
```diff
@@ -1,21 +1,86 @@
...
+              }
 
-}
+              if (atm.sumChecking < 0 && atm.sumSavings >= 0) {
+                if (atm.sumSavings <= 0) {
+                  $( "#savings .balance").addClass( "zero");
+                  atm.sumChecking = 0;
+                  return alert("You don't have enough money for this transaction.")
+                }
+                // Overdraft amount ( ex. -10 + 50 = 40)
```
jshawl, line 39: helpful code comments!

jshawl:
:+1: :star: excellent job on this assignment! You might want to think about ways to combine the functionality between the checking and savings accounts so you don't need separate code for each account. i.e. what if the user wanted to add another checking account? it would be nice if you didn't have to hardcode a new function for each new account.

PR #17, opened 2015-07-10
Alexbannon solution

alexbannon:
- comfort: 5/5
- completeness: 5/5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,15 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$("input").eq(1).on("click", depositIntoChecking);
+$("input").eq(4).on("click", depositIntoSavings);
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
+function depositIntoChecking(event) {
+  event.preventDefault();
+  var checkingAccountUserInput = Number(parseFloat($("input").eq(0).val()).toFixed(2));
+  var checkingTotal = (parseFloat($("div.balance").eq(0).html().substr(1))) + checkingAccountUserInput;
+  $("div.balance").eq(0).html("$" + checkingTotal);
 }
-
-function checkingWithdrawal(amount){
-
+function depositIntoSavings(event) {
+  event.preventDefault();
+  var savingsAccountUserInput = Number(parseFloat($("input").eq(3).val()).toFixed(2));
+  var savingsTotal = (parseFloat($("div.balance").eq(1).html().substr(1))) + savingsAccountUserInput;
+  $("div.balance").eq(1).html("$" + savingsTotal);
 }
-
-function savingsDeposit(amount){
-
-}
-
-function savingsWithdrawal(amount){
-
-}
-
-// an eventListerner for each button, each one a "click"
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,96 @@
...
+});
+$("input").eq(3).on("keypress", function(event) {
+  if(event.which == 13) {
+    event.preventDefault();
+    var savingsAccountUserInput = Number(parseFloat($("input").eq(3).val()));
+    var savingsTotal = (parseFloat($("div.balance").eq(1).html().substr(1))) + savingsAccountUserInput;
+    $("div.balance").eq(1).html("$" + savingsTotal.toFixed(2));
+    $("input").eq(3).val("");
+    $(".account").eq(1).removeClass("zero");
+  }
```
adambray, line 25: This code looks almost exactly the same as the function `depositIntoSavings`. Duplicated code is very undesirable in a program, so I would encourage you to see if you could find a way to re-use the function you wrote below (maybe just call it inside your `if(event.which === 13)` block?

atm.js
```diff
@@ -1,21 +1,96 @@
...
 }
-
-function savingsDeposit(amount){
-
+function withdrawFromChecking(event) {
+  event.preventDefault();
+  var checkingAccountUserWithdrawal = Number(parseFloat($("input").eq(0).val()));
+  var checkingTotalAfterWithdrawal = (parseFloat($("div.balance").eq(0).html().substr(1))) - checkingAccountUserWithdrawal;
+  var savingsTotal = (parseFloat($("div.balance").eq(1).html().substr(1)));
+  if(checkingTotalAfterWithdrawal + savingsTotal < 0) {
```
adambray, line 48: Using blank lines (whitespace) can make your code much more readable. For example, I'd include a blank line above this if statement.

atm.js
```diff
@@ -1,21 +1,96 @@
...
+  $("div.balance").eq(1).html("$" + savingsTotal.toFixed(2));
+  $("input").eq(3).val("");
+  $(".account").eq(1).removeClass("zero");
 }
-
-function savingsDeposit(amount){
-
+function withdrawFromChecking(event) {
+  event.preventDefault();
+  var checkingAccountUserWithdrawal = Number(parseFloat($("input").eq(0).val()));
```
adambray, line 45: I'd make this variable name a bit shorter, but more descriptive if possible, something like `amountToWithdraw`.

atm.js
```diff
@@ -1,21 +1,96 @@
...
+  var checkingTotalAfterWithdrawal = (parseFloat($("div.balance").eq(0).html().substr(1))) - checkingAccountUserWithdrawal;
```
adambray, line 46: I'd also extract out the complex part here into another variable, something like:

``` js
var currentBalance = parseFloat($("div.balance").eq(0).html().substr(1));
var checkingTotalAfterWithdrawal = currentBalance - checkingAccountUserWithdrawal;
```

atm.js
```diff
@@ -1,21 +1,96 @@
...
+      $(".account").eq(1).addClass("zero");
+    }
+    $("div.balance").eq(1).html("$" + savingsTotal.toFixed(2));
+    $("input").eq(0).val("");
+  }
 }
-
-function savingsWithdrawal(amount){
-
+function withdrawFromSavings(event) {
```
adambray, line 70: This function is also almost entirely duplicated compared to the function above, only the two accounts are switched. Wonder if you could re-write as one function, that takes two inputs, it might look like this:

``` js
function withdrawFunds(amount, primaryAccount, overdraftAccount) {
  // your code here
}

withdrawFunds(100, checkingAccount, savingsAccount);
```

adambray:
Good job! I know this is a tough problem, but you solved it well! Just keep trying to find ways to shorten / simplify your code, and let me know if you have any questions.

PR #18, opened 2015-07-10
Erica Irving solution

ebirving:
comfort_level: 4
completeness: 5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,12 +1,12 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+//var checkingBalance = document.querySelector("div#checking_balance");
+var checkingBalance = parseInt(document.querySelector("div#checking_balance").innerHTML.replace("$", ""));
 
 //Instantiate var savingsBalance in the style of var checkingBalance
-var savingsBalance = document.querySelector("div#savings_balance");
-savingsBalance = parseInt( savingsBalance.innerHTML.replace("$", "") );
+//var savingsBalance = document.querySelector("div#savings_balance");
+savingsBalance = parseInt(document.querySelector("div#savings_balance").innerHTML.replace("$", ""));
 
 var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+  // checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,71 @@
...
 
-}
+$("input[value='Deposit']").eq(0).on("click", function(){
+    var deposit = parseFloat($("input#checking_input").val());//Set their input equal to amount
+    checkingBalance += deposit;//Add amount to checkingBalance, i.e. checkingBalance += amount
+    $("div#checking_balance").html("$ " + checkingBalance.toFixed(2))//Update the "balance" div to reflect the new checkingBalance
+    if(checkingBalance > 0) {
+      $("div.account").eq(0).removeClass("zero");//Remove class zero
+    };
+})
```
andrewsunglaekim, line 15: I think you might be repeating this code where we remove or add class zero based on balance. It might be better to abstract this into a function so you can just call the function any time you need to set classes.

atm.js
```diff
@@ -1,21 +1,71 @@
...
+      if(((checkingBalance + savingsBalance) - withdrawal) > 0){
+        savingsBalance -= (withdrawal - checkingBalance);
+        $("div#savings_balance").html("$ " + savingsBalance.toFixed(2));//Update the savings balance div
+        checkingBalance = 0;
+        $("div.account").eq(0).addClass("zero");
+      }
+      else {
+        alert("Insufficient funds");
+      }
+    }
```
andrewsunglaekim, line 30: noice over draft protection!

atm.js
```diff
@@ -1,21 +1,71 @@
...
+    else {
+      checkingBalance -= withdrawal;
+      if (checkingBalance == 0) {
+        $("div.account").eq(0).addClass("zero");
+      }
+    }
+  $("div#checking_balance").html("$ " + checkingBalance.toFixed(2));//Update the "balance" div to reflect the new checkingBalance
+})
```
andrewsunglaekim, line 38: i noticed you used `parseFloat()` above and is subsequently why you have to use `.toFixed` here I think this way is fine. But for the scope of this problem it may be easier to user `parseInt()` I think either way works. In fact your way is better because it incorporates cents... now that i think about it.

andrewsunglaekim:
Overall, really great job erica! I think there are some refactoring opportunities here though. I think you may be able to abstract some of this code out into different functions and call them in lieu of stringing together all of your functionality/jquery. Not a necessity but a nice to have. This will become clearer later in the course when we start executing object oriented approaches to our code.

PR #19, opened 2015-07-10
Coulter AMT HW

kscoulter:
confidence: 3.5;
completeness: 4;

I got my code to work first, then I decided to refactor it to be object-oriented. I didn't have time to finish overdraft protection or input validation. I'm still learning ins and outs of object. I would love for the next review to go over objects again.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,47 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+//Checking account
+//Define players
+$(document).ready(function() {
+  var accountBalance = 0;
+  var inputBox = $("input:first")[0];
+  var userAmount = $("input:first").val();
+  var depositButton = $("input").eq(1);
+  var withdrawButton = $("input").eq(2);
+  //$("input").eq(1).attr("value") is "Deposit"
+  //$("input").eq(2).attr("value") is "Withdraw"
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+  //When Deposit button is clicked, read user input and add it to the total
+  depositButton.on("click", function(){
+    event.preventDefault();
+  })
 
-}
 
-function checkingWithdrawal(amount){
+//Whin withdraw button is clicked, read user input and subtract it from the total
+//add logic to make sure the balance is not zero
 
-}
+// var checkingBalance = document.querySelector("div#checking_balance");
+// checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+//
+// var checkingDeposit = function(amount){
+//   checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+//
+// }
+//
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,54 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
+//Checking account
+//Define players
+$(document).ready(function() {
```
jshawl, line 3: this is not wrong, but `document.ready` is unecessary here because your script tag is just before the closing `body` tag, and the dom will have loaded by the time the script loads.

atm.js
```diff
@@ -1,21 +1,54 @@
...
+
+var atm = {
```
jshawl, line 5: Excellent use of OOJS!!!! :) :star2:

atm.js
```diff
@@ -1,21 +1,54 @@
...
+      event.preventDefault();
+      var actionType = $(event.target).val();//Deposit of withdraw
+      var accountType = $(event.target).prevAll("h2").html();//Checking or savings
+      var amount = parseInt($(event.target).prevAll("[type=text]").val());//amount
+      if( actionType === "Deposit"){
+        this.deposit(accountType,amount)
+      }
+      if( actionType === "Withdraw"){
+        this.withdraw(accountType,amount)
+      }
```
jshawl, line 21: or `actionType === "Deposit" ? this.deposit(accountType,amount) : this.withdraw(accountType,amount)`

turnary operators are cool but harder to read. just wanted to share!

atm.js
```diff
@@ -1,21 +1,54 @@
...
+    }.bind(this))
+  },
+  deposit: function(account,amount){
+    account = account.toLowerCase();
+    this.accounts[account] += amount;
```
jshawl, line 26: YES ^^^^ !!!!

atm.js
```diff
@@ -1,21 +1,54 @@
...
+      this.accounts[account] -= amount;
+      $(event.target).prevAll("[type=text]").val(""); //empty the form after submission
+      this.updateDisplay();
+      console.log("savings balance is " + atm.accounts.savings);
+      console.log("checking balance is " + atm.accounts.checking);
+    }
+  },
+  updateDisplay: function(){
+    $(".balance").eq(0).html("$" + atm.accounts.checking);
+    $(".balance").eq(1).html("$" + atm.accounts.savings);
```
jshawl, line 44: Can you think of a way to avoid hardcoding which accounts need to be displayed? maybe a `for .. in` loop?

``` js
for( var account in atm.accounts ){
  $(".balance + ???").html("$" + account)
}
```

jshawl:
:+1: :star: I am very impressed with your OOJS solution! I would encourage you to think about how you can split your applications into two - the engine (the application logic) and the view (the code responsible for representing the data)

atm.js
kscoulter, line 5: Thanks, one-on-one with you really helped!

PR #21, opened 2015-07-10
ch-ching!

jkoul:
Comfort: 4
Completeness: 5

This assignment was really useful. It forced me to get a much stronger handle on both the interaction of jQuery with event listeners and the challenges resulting from scope. So while I ended up as a 4 on comfort, I definitely didn't start there.

At the same time, I know my code has the potential to be quite a bit tighter, which is why I only gave myself a 4 on comfort. I kept running into roadblocks when specifying the functions resulting from clicking the four buttons. I also was forced to edit the DOM selectors in the HTML file, which I had hoped to avoid. But overall, I feel like I ultimately handled this one pretty well.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,84 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var chkDeposit = $("#checking_deposit");
+var chkWithdrawal = $("#checking_withdraw");
+var savDeposit = $("#savings_deposit");
+var savWithdrawal = $("#savings_withdraw");
+var chkAmount = 0
+var savAmount = 0
 
+
+// an eventListener for each button, each one a "click"
+chkDeposit.on("click", checkingDeposit);
+chkWithdrawal.on("click", checkingWithdrawal);
+savDeposit.on("click", savingsDeposit);
+savWithdrawal.on("click", savingsWithdrawal);
+
+//actions responding to each button
+function checkingDeposit() {
+  chkAmount = parseInt($("#checking_amount").val());
+  var checkingBal = $(".balance.checking").html();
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,84 @@
...
+var savAmount = 0
 
-}
 
-function checkingWithdrawal(amount){
+// an eventListener for each button, each one a "click"
+chkDeposit.on("click", checkingDeposit);
+chkWithdrawal.on("click", checkingWithdrawal);
+savDeposit.on("click", savingsDeposit);
+savWithdrawal.on("click", savingsWithdrawal);
```
jshawl, line 14: nice! can you think of a way to combine the event listeners into one? maybe you can check the text of the button to decide which function to call

atm.js
```diff
@@ -1,21 +1,84 @@
...
 
-}
+//actions responding to each button
+function checkingDeposit() {
+  chkAmount = parseInt($("#checking_amount").val());
+  var checkingBal = $(".balance.checking").html();
+  var chkBalAmt = parseInt(checkingBal.replace("$", ""));
+  checkingBal = chkBalAmt + chkAmount;
+  $(".balance.checking").text("$" + checkingBal);
+  if(checkingBal == 0) {
```
jshawl: or just `if(checkingBal)`, because `0` is falsey in JavaScript

atm.js
```diff
@@ -1,21 +1,84 @@
...
+  };
+};
 
-function savingsDeposit(amount){
 
+function checkingWithdrawal(){
+  chkAmount = parseInt($("#checking_amount").val());
+  var checkingBal = $(".balance.checking").html();
+  var chkBalAmt = parseInt(checkingBal.replace("$", ""));
+  var savingsBal = $(".balance.savings").html();
```
jshawl, line 35: Use variable names when selecting elements with jQuery to increase your app's performance.

jshawl:
:+1: Really excellent job on this assignment! If you wanted to keep working on this, I would encourage you to separate your application's logic from the way it's displayed on the page. That is, you current need separate functions for each account. Can you think of a way to organize your code so that the user could have an arbitrary amount of accounts without having to write more javascript?

PR #23, opened 2015-07-10
finished atm hw

adrianshep:
comfort_level = 2 at start, rose to 4 with Andy's review and example
completeness = 5 (not enough time left to figure out the overdraft)

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,87 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready(function(){
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+    $(".checkDepBtn").on("click", function(){
+      var deposit = parseInt($(".checkEntry").val());
+      var balance = $("#checkDisplay").text()
+      balance = parseInt(balance.replace("$", ""))
+      var checkingBalance = (deposit + balance)
+      $("#checkDisplay").text("$" + checkingBalance)
+    })
 
-}
+    $(".checkWdrwBtn").on("click", function(){
+      var withdrawal = parseInt($(".checkEntry").val())
+      var balance = $("#checkDisplay").text()
+      balance = parseInt(balance.replace("$", ""))
+      var checkingBalance = (balance - withdrawal)
+        if (checkingBalance >= 0) {
+          if (checkingBalance !== 0) {
+            $("#checkDisplay").text("$" + checkingBalance)
+            }
+          else {
+            $("#checkDisplay").text("$" + checkingBalance)
+            $("#checkDisplay").css("background-color", "#F52F4F");
+          }
+        }
+      })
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,87 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready(function(){
```
jshawl, line 1: this is not wrong, but `document.ready` is unecessary here because the script tag is just before the closing body tag. The DOM will have loaded by the time your script is called.

atm.js
```diff
@@ -1,21 +1,87 @@
...
+      var balance = $("#checkDisplay").text()
+      balance = parseInt(balance.replace("$", ""))
+      var checkingBalance = (balance - withdrawal)
+        if (checkingBalance >= 0) {
+          if (checkingBalance !== 0) {
+            $("#checkDisplay").text("$" + checkingBalance)
+            }
+          else {
+            $("#checkDisplay").text("$" + checkingBalance)
+            $("#checkDisplay").css("background-color", "#F52F4F");
```
jshawl, line 22: I recommend toggling a class here instead of setting an inline style. If you toggle a class, you can separate your concerns by keeping behavior in your JS files and styles in your CSS files.

atm.js
```diff
@@ -1,21 +1,87 @@
...
-}
+      $(".saveWdrwBtn").on("click", function(){
+        var withdrawal = parseInt($(".saveEntry").val())
+        var balance = $("#saveDisplay").text()
+        balance = parseInt(balance.replace("$", ""))
+        var savingsBalance = (balance - withdrawal)
+          if (savingsBalance >= 0) {
+            if (savingsBalance !== 0) {
+              $("#saveDisplay").text("$" + savingsBalance)
+              }
```
jshawl, line 43: This code looks very similar to the code above for the checking withdrawal. Can you think of a way to share the functionality between the two accounts?

atm.js
```diff
@@ -1,21 +1,87 @@
...
+            else {
+              $("#saveDisplay").text("$" + savingsBalance)
+              $("#saveDisplay").css("background-color", "#F52F4F");
+            }
```
jshawl, line 47: it would be nice to alert the user if they didn't have enough money to perform a transaction. Currently your application will fail silently.

atm.js
```diff
@@ -1,21 +1,87 @@
...
 
-// an eventListerner for each button, each one a "click"
\ No newline at end of file
+//overdraft mechanism for Checking and Savings
+// = if transaction would send active account below zero,
+//    and if it would send inactive account balance below zero,
+//    don't transact; else if inactive account has sufficient balance to
+//    cover active account transaction, adjust active account balance to
+//    zero and take away difference of transaction from inactive account
+//    and adjust inactive account balance.
```
jshawl, line 87: Excellent pseudo code! When I pseudocode, I like to leave the comments in place and write the corresponding code below each line

jshawl:
:+1: Excellent job on this assignment!

PR #25, opened 2015-07-10
Erica Andersen's ATM Solution

ericacomfort:
Comfort: 3.5/5
Completeness: 4/5
Concerns: I really need to work on DRYing up my code. Also, I know there's a much better way to add '.zero' but I just didn't quite have the time to think it through.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -2,24 +2,33 @@
 var balanceDisplay = $(".account").children().eq(1).html(); // Selects Balance HTML Area (cannot be called in function?)
 var userInput = $(".account").children().eq(2); // Selects User Input
 var checkDepositButton = $(".account").children().eq(3); // Checking Deposit Button
+var checkWithdrawlButton = $(".account").children().eq(4); // Checking Withdrawl Button
 
 $(checkDepositButton).click(function() { // Set up Checking Deposit button
   event.preventDefault(); // Prevents Default Refresh
-  var depAmount = parseInt($( userInput ).val()); // New Total Value
-  var balance = parseInt($( "#check-balance" ).text().replace("$", "")) // Does not work with stored variable
-  var newBalance = "$" + (balance + depAmount)
-  $( "#check-balance" ).html(newBalance); // Does not work with stored variable
+  var depAmount = parseInt($( userInput ).val()); // Get Deposit Value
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,87 @@
...
+var userInputChecking = $(".account").children().eq(2); // Selects Checking User Input
+var checkDepositButton = $(".account").children().eq(3); // Checking Deposit Button
+var checkWithdrawlButton = $(".account").children().eq(4); // Checking Withdrawl Button
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var balanceDisplaySavings = $(".account").children().eq(6); // Selects Savings Balance HTML Area
+var userInputSavings = $(".account").children().eq(7); // Selects Savings User Input
+var savingsDepositButton = $(".account").children().eq(8); // Savings Deposit Button
+var savingsWithdrawlButton = $(".account").children().eq(9); // Savings Withdrawl Button
```
adambray, line 10: I like how you defined all of your UI variables up at the top, and in a neat and organized way! One small suggestion... since you can edit this HTML, why not add ids or classes so that you can use those in your selectors instead of `.children().eq(x)`; this would make your code more readable, and less likely to break if the HTML is changed.

e.g. `$(".account button#deposit-btn")`

atm.js
```diff
@@ -1,21 +1,87 @@
...
+  }
+  else if((totalBalance - withAmount) >= 0){ // Overdraft only if Total Value is enough
+    var balanceOverdraft = withAmount - balanceChecking; // Find how much can be withdrawn
+    var newbalanceSavings = "$" + (balanceSavings - balanceOverdraft); // Subtract Overdraft Value from Savings Value
+    balanceDisplay.html(newbalanceSavings); // Set Checking Balance to New Balance after Overdraft
+    balanceDisplaySavings.html("$0"); // Set Savings Balance to '$' When Overdraft happens
+  }
+  console.log("Check Withdrawl button is working")
+  console.log("New Savings Balance is: " + newbalanceSavings)
+});
```
adambray, line 49: Good job keeping this neat and well-commented!

adambray:
Good job! :+1: 

Not many suggestions for improvement, except that there is a bit of duplication in your deposit code for checking and savings... it would be a good exercise to find a way to wrap that code up into a function and re-use it for both accounts.

PR #28, opened 2015-07-10
Almost completely finished ATM

walter-0:
Comfort 4/5
Completeness 4.5/5
 almost completely made to specs. the only thing missing is preventing a negative value in the savings account when the checking account is overdrawn

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,51 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$( document ).ready(function() {
+    console.log( "ready!" );
+});
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-}
+//add value of inputChecking to #checking
+$("#checkingDeposit").click(function(){
+  var inputChecking = parseInt($("#inputChecking").val());
+  var checking = $("#checking").text();
+  checking = parseInt(checking.replace("$", ""))
+  var newBalance = checking + inputChecking
+  console.log(newBalance);
+  $("#checking").text("$" + newBalance)
+})
 
-function checkingWithdrawal(amount){
+//subtract inputChecking from #checking
+$("#checkingWithdrawal").click(function(){
+  var inputChecking = parseInt($("#inputChecking").val());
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,74 @@
...
+    var checking = parseInt($("#checking").text().replace("$", ""));
+    var newBalance = checking + inputChecking
+    console.log(newBalance);
+    $("#checking").text("$" + newBalance)
+    if (checking > 0) {
+      $(".checkingAccount").removeClass("zero")
 
-function checkingWithdrawal(amount){
+    }
+  })
```
adambray, line 17: Well done here! I like the readable variable names. :+1:

atm.js
```diff
@@ -1,21 +1,74 @@
...
 
-}
+  //subtract inputChecking from #checking
+  $("#checkingWithdrawal").click(function(){
+    var inputChecking = parseInt($("#inputChecking").val());
```
adambray, line 21: I might call this var something more descriptive, like `amountToWithdraw`

atm.js
```diff
@@ -1,21 +1,74 @@
...
+      var overDraft = inputChecking - checking
+      $("#checking").text("$0")
+      var newSavingsBalance = (savings - overDraft)
+      $("#savings").text("$" + newSavingsBalance)
+      $(".checkingAccount").addClass("zero")
+    }
 
-}
+    //I could not get the savings to prevent a negative balance when overdraft
+    else if ((newBalance <= 0) || (newSavingsBalance <= 0)){
```
adambray, line 35: 2 things:
1. This shouldn't be in an else condition, because that means it will only run if the associated if (line 26 above) is false (i.e. if there was no need to overdraft). Another way to say this is that this bit of code (L35 - 37) should be nested inside the previous if statement.
2. You should only need to check if newSavingsBalance is less than 0 in that case.

atm.js
```diff
@@ -1,21 +1,74 @@
...
+    if (newBalance <0) { //prevents negative balances
+      return;
+    }
+    console.log(newBalance);
+    $("#savings").text("$" + newBalance)
+    if (savings = 0) {
+      $(".savingsAccount").removeClass("zero")
+    }
+  })
+});
```
adambray, line 74: These two functions look good!

adambray:
Good job! Let me know if you have any questions about my comments.
:heavy_check_mark:

PR #29, opened 2015-07-10
Atm HW

omarimayerswalker:
Understanding: 4
Completeness: 5

This one gave me trouble at first. I had to scrap my first try and simplify. I wasn't able to figure out how to create the overdraft function.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -48,6 +48,18 @@ function deposit (deposit){
     return currentCheckBalance || currentSaveBalance;
 }
 
+function withdraw (withdraw){
+    if (withdraw == checkingInput){
+        currentCheckBalance = currentCheckBalance - parseInt(withdraw);
+    }
+
+    else if (withdraw == savingsInput){
+        currentSaveBalance = currentSaveBalance - parseInt(withdraw);
+        console.log('withdraw output');
+    }
+    return currentCheckBalance && currentSaveBalance;
+}
+
 // balance window function
 
 function balanceWindow (accountBalance){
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,77 @@
...
+var currentSaveBalance = 0;
 
-}
+// event listeners
 
-function checkingWithdrawal(amount){
+// checking deposit button
 
-}
+$('#checkDepositButton').click('click', function(){
```
jshawl, line 10: use `.on("click")` instead of `.click("click")`. I wonder what other arguments you can pass to `.click?`

atm.js
```diff
@@ -1,21 +1,77 @@
...
+    var deposit = $('#checkingInput').val();
+    currentCheckBalance = currentCheckBalance + parseInt(deposit);
+    $('#checkBalanceWindow').text('$' + currentCheckBalance);
 
-function savingsDeposit(amount){
+    // zero balance color change
+    if (currentCheckBalance == 0){
+        $('#checkBalanceWindow').css('background-color', 'red');
```
jshawl, line 17: I recommend toggling a class here instead of setting inline styles. If you toggle a class, you can separate your concerns by keeping behavior in JS files and styles in CSS files.

atm.js
```diff
@@ -1,21 +1,77 @@
...
 
-}
+// withdraw deposit button
+
+$('#checkWithdrawButton').click('click', function(){
+    var withdraw = $('#checkingInput').val();
+    if (currentCheckBalance >= withdraw){
+        currentCheckBalance = currentCheckBalance - parseInt(withdraw);
+        $('#checkBalanceWindow').text('$' + currentCheckBalance);
+    }
```
jshawl, line 31: You mentioned having trouble with the overdraft protection - but I don't see any code here for it. Let me know if you'd like to go over the overdraft solution in a one on one!

jshawl:
:+1: Nice job on this assignment! I would encourage you to think of a way to combine the functionality between the two accounts into a function or two. That is, if you wanted users to be able to have any number of accounts - what would the javascript for that look like?

PR #30, opened 2015-07-10
in jquery, works without overdraft protection

LauraWN:


mattscilipoti:
:+1: Looks good.  Thanks for the comments and organization.

Next time, after you get it to work, see if you can remove some duplication.

PR #31, opened 2015-07-10
Stephen Gheysens ATM Solution

segheysens:
Comfort: 4/5
Completeness: 5/5

I think the only question I had was if in the below code: 

``` javascript
$("[value=\"Deposit\"]").eq(0).click( function(event){checkingDeposit(event);} );
```

I had just written the checkingDeposit function like so:

``` javascript
$("[value=\"Deposit\"]").eq(0).click( checkingDeposit(event) );
```

... why would it get invoked? Does the first one not get invoked because it's encapsulated in the anonymous function? It just seemed to me that if the second version was invoking, the first would invoke the anonymous function, and subsequently invoke the checkingDeposit function. Thanks in advance for your help!

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -7,22 +7,26 @@ var savingsInt = Number.parseInt(savingsBalance.innerHTML.replace("$", ""));
 
 
 // add eventListener to each button
-$("[value=\"Deposit\"]")[0].click( checkingDeposit(event) );
-$("[value=\"Withdrawal\"]")[0].click( checkingWithdrawal(event) );
+$("[value=\"Deposit\"]").eq(0).click( function(event){checkingDeposit(event);} );
+$("[value=\"Withdraw\"]").eq(0).click( function(event){checkingWithdrawal(event);} );
 
 // event listeners for savings account buttons
-$("[value=\"Deposit\"]")[1].click( savingsDeposit(event) );
-$("[value=\"Withdrawal\"]")[1].click( savingsWithdrawal(event) );
+$("[value=\"Deposit\"]").eq(1).click( function(event){savingsDeposit(event);} );
+$("[value=\"Withdraw\"]").eq(1).click( function(event){savingsWithdrawal(event);} );
 
 
 function checkingDeposit(event){
   // Check contents (integer form) of text input field
+  checkingInt = Number.parseInt(checkingBalance.innerHTML.replace("$", ""));
+  console.log("CHECKING DEPOSIT")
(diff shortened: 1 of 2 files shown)


andrewsunglaekim:
So, when we pass in functions as a callback, without parens references them and with parens invokes. When we pass in an anonymous function it doesn't get called until execution of the callback. If we pass in the function itself it gets not only referenced but invoked at that same moment

atm.js
```diff
@@ -1,21 +1,118 @@
...
+  // Sum together that number and the number parsed from the current balance
+  checkingBalance.innerHTML = "$" + (checkingInt + Number.parseInt($("[type=\"text\"]").eq(0).val()));
+  // Clear the text input field
+  $("[type=\"text\"]").eq(0).val("");
 
+  // if the checking account balance > 0, add the class 'zero'
+  if(Number.parseInt(checkingBalance.innerHTML.replace("$", "")) > 0){
+    $(".balance").eq(0).removeClass("zero");
+  }
 }
```
andrewsunglaekim, line 37: might be good to encapsulate in this method, because you will be repeating this functionality any time you change the balance. Additionally, i think its only necessary for the withdrawals. Although, it would be modular to have it call this function any time we check the balance. Maybe we want to do other stuff when we check the balance instead of just adding or removing the zero class going forward.

atm.js
```diff
@@ -1,21 +1,118 @@
...
+  savingsInt = Number.parseInt(savingsBalance.innerHTML.replace("$", ""));
+  // if withdraw > checkingBalance, check if the remaining withdrawal request could be taken from savingsBalance
+  if (Number.parseInt($("[type=\"text\"]").eq(0).val()) > checkingInt){
+    console.log("if 1");
+    //  if remainder can be taken from savingsBalance, make checkingBalance zero then take the remainder from savingsBalance
+    if(savingsInt - (Number.parseInt($("[type=\"text\"]").eq(0).val()) - checkingInt) >= 0){
+      console.log("if 2");
+      checkingBalance.innerHTML = "$0";
+      savingsBalance.innerHTML = "$" + (savingsInt - (Number.parseInt($("[type=\"text\"]").eq(0).val()) - checkingInt));
+    }
```
andrewsunglaekim, line 58: :+1:  i like the overdraft protection!

andrewsunglaekim:
Nice job overall stephen! I think there's a couple of refactor opps here and there, but nothing too crazy. Let me know if you have any questions.

PR #32, opened 2015-07-10
ATM hw

ghost:
comfort: 4
completeness: 4.5

All buttons, calculations, and overdraft protection are working. 

I still don't get the color change using removeClass() and addClass().....

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,7 +1,7 @@
 $(document).ready(function(){
 
-  checkingAmount = $("#checkingBalance");
-  savingAmount = $("#savingBalance");
+  checkingAmount = $(".balance.checking");
+  savingAmount = $(".balance.saving");
 
   checkingInput = $("input.checkingInput");
   savingInput = $("input.savingInput");
@@ -28,6 +28,7 @@ $(document).ready(function(){
       $("div#checkingBalance").text("$" + newAmount);
 
       checkingInput.val("");
+
     }
 
     function withdrawCheckingMoney(){
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,83 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var bankAccount = {
+  elements: {
+    checkingDepositButton: document.getElementsByTagName("input")[1],
+    checkingWithdrawButton: document.getElementsByTagName("input")[2],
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,101 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready(function(){
```
jshawl, line 1: This works but is unnecessary, as the DOM will have loaded by the time this script is called just before the closing body tag.

atm.js
```diff
@@ -1,21 +1,101 @@
...
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+  checkingAmount = $(".balance.checking");
```
jshawl, line 3: remember to use `var` to prevent global variables

atm.js
```diff
@@ -1,21 +1,101 @@
...
+  savingDeposit = $(".savingDeposit");
+  savingWithdraw = $(".savingWithdraw");
 
-}
+  checkingDeposit.on("click", depositCheckingMoney);
+  checkingWithdraw.on("click", withdrawCheckingMoney);
 
-function savingsDeposit(amount){
+  savingDeposit.on("click", depositSavingMoney);
+  savingWithdraw.on("click", withdrawSavingMoney);
```
jshawl, line 18: Can you think of a way to use one listener for each of these click handlers? i.e. maybe you can check the text of the button that's clicked on to decide what to do.

atm.js
```diff
@@ -1,21 +1,101 @@
...
+      var withdraw = parseInt(checkingInput.val());
+
+      var savingCurrentBalance = $("#savingBalance").text();
+      savingCurrentBalance = parseInt(savingCurrentBalance.replace("$", ""));
+
+      var overDraftProtection = checkingCurrentBalance + savingCurrentBalance - withdraw;
+
+        if (checkingCurrentBalance >= withdraw){
+          var sum = checkingCurrentBalance - withdraw;
+          $("div#checkingBalance").text("$" + sum);
```
jshawl, line 48: Try saving jQuery Selectors to a variable so you don't have to repeat yourself

jshawl:
:+1: Looks good Tae! Let me know if you'd like to work on `removeClass()` and `addClass()` in our one on one.

PR #33, opened 2015-07-10
Mostafa Habib ATM Solution

M00staff:
comfort 4
completeness 4

i got rid of the existing code because it was throwing me off.  I also had to add IDs for the buttons and text in the HTML for simplicity's sake.  I didn't get a chance to add in the NaN error or get it to turn back from red to white when you go back into the positive.  Or the $ sign - but all that was from a lack of time, as I'm pretty sure all it would require are some if/then statements and a little concatenating.   Oh and the pulling from savings - again time.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,50 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+//my likely bad code
+// var checkBal = $('.balance').eq(0).html();
+// var checkSave = $('.balance').eq(1).html();
 
+
+var checkBalance = 0;
+
+//left side number, currently null
+// var checkingBalance = document.querySelector("div#checking_balance");
+//     checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+
+
+//right side number in a varible and function
 var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+    checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-}
+    }
 
 function checkingWithdrawal(amount){
-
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,63 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkBalance = 0;
+var saveBalance = 0;
```
andrewsunglaekim, line 3: I think going forward, we'll be persisting data. Potentially we could have a balance of 150, but with your application it would automatically update the balance from 0.

atm.js
```diff
@@ -1,21 +1,63 @@
...
+  userLeftAmount = $('#leftAmount').val();
+  console.log(userLeftAmount);
+  checkBalance = parseInt(checkBalance) - parseInt(userLeftAmount);
+  console.log(checkBalance);
+  $('.balance').eq(0).html(checkBalance);
 
-}
+  if (checkBalance < 0){
+  $('.balance').eq(0).html('you broke');
+  $('.balance').css("background-color" , "red");
```
andrewsunglaekim, line 26: noice

atm.js
```diff
@@ -1,21 +1,63 @@
...
+
+  if (saveBalance < 0){
+  $('.balance').eq(1).html('you broke');
+  $('.balance').css("background-color" , "red");
+  }
+  else{
+  $('.balance').eq(1).html(saveBalance);
+  }
+
+});
```
andrewsunglaekim, line 63: I like how you logged alot of your variables.. Couple of things: when you do things for realisies you might wanna take them out when you go to production. Additionally you may want to be more semantic, ie say the action thats happening when the log happens.

andrewsunglaekim:
nice work on this assignment moose. How might we go about implementing over drawing from an account? Let me know if you have any questions.

PR #35, opened 2015-07-10
ATM homework

chrislovett03:
Completeness: 2
Comfort: 2.5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,45 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingWithdrawButton = $("#withdrawButton");
+var checkingBalance = $("#checkingBalance");
+var checkingNewDeposit = $("#valueinput");
+var checkingValue = $("#checkingNewDeposit.val()");
+var cBalance= checkingBalance.html().substr(1);
+var checkingDespositButton = $("#depositButton");
+checkingDespositButton.on("click", function(){
+  var checkingNewDeposit = $("#valueinput");
+  var checkingTotal = parseInt(checkingValue) + parseInt(cBalance);
+  var cBalance= checkingBalance.html().substr(1);
+    console.log(checkingTotal)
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,45 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
+var checkingWithdrawButton = $("#withdrawButton");
+var checkingBalance = $("#checkingBalance");
+var checkingNewDeposit = $("#valueinput");
+var checkingValue = $("#checkingNewDeposit.val()");
+var cBalance= checkingBalance.html().substr(1);
```
jshawl, line 5: This is repeated below - it has to be inside the click handler so that the variable is stored after the balance is updated

atm.js
```diff
@@ -1,21 +1,45 @@
...
+var checkingDespositButton = $("#depositButton");
+checkingDespositButton.on("click", function(){
+  var checkingNewDeposit = $("#valueinput");
+  var checkingTotal = parseInt(checkingValue) + parseInt(cBalance);
+  var cBalance= checkingBalance.html().substr(1);
+    console.log(checkingTotal)
```
jshawl, line 11: excellent job on indentation on this assignment! there are just a few `console.log`s that should be aligned with the rest of the code.

atm.js
```diff
@@ -1,21 +1,45 @@
...
+    checkingBalance.html("$" + checkingTotal)
+})
+checkingWithdrawButton.on("click", function(){
+  var checkingTotal = parseInt(checkingValue) - parseInt(cBalance);
+    console.log(checkingTotal)
+    checkingBalance.html("$" + total)
+})
```
jshawl, line 18: Excellent Job on this assignment Christine!!! :star2: The code below looks very similar to the code above. Can you think of a way to handle both accounts without two separate functions and click handlers? i.e. what if you wanted the user to be able to create another checking or savings account?

jshawl:
:+1:

PR #37, opened 2015-07-10
Hw Submission

edwardpark:
comfort:4 ( until I started refactoring to OO therein it became a 2.223)
complete: 5

### DIFF
--- a/atm.css
+++ b/atm.css
@@ -18,7 +18,7 @@ body {
   font-size: 50px;
 }
 
-.account {
+.checking_account, .savings_account {
   background-color: #6C9A74;
   border: 1px solid #295A33;
   border-radius: 15px;
--- a/atm.js
+++ b/atm.js
@@ -1,6 +1,8 @@
 
 var checkingBalance = 0;
 var savingsBalance = 0;
+
+// CHECKING ACCOUNT
 var checkingDisplay = $(".checking_account .balance");
 var checkingInput = $(".checking_input");
 function checkingDeposit(){
@@ -11,7 +13,6 @@ function checkingDeposit(){
   checkingDisplay.css("background-color","#61c419");
 
 
-  //target and get value inside text box
 }
 
 function checkingWithdrawal(){
@@ -23,10 +24,12 @@ function checkingWithdrawal(){
 
 
 }
-//
+
+ //SAVINGS ACCOUNT
 
 var savingsDisplay = $(".savings_account .balance");
 var savingsInput = $(".savings_input");
+
 function savingsDeposit(){
   var depositAmount = parseFloat($(savingsInput).val());
   savingsBalance += depositAmount;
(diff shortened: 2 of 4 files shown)


atm.css
```diff
@@ -18,7 +18,7 @@ body {
   font-size: 50px;
 }
 
-.account {
+.checking_account, .savings_account {
```
mattscilipoti, line 21: Since this is common to both, it may make sense to leave the "account" class, but add an additional class (or ID) for specificity in other contexts.

atm.js
```diff
@@ -1,21 +1,101 @@
...
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+function overDraft(){
+  if (savingsBalance > 0) {
+    var depositAmount = parseFloat($(checkingInput).val());
+    savingsBalance -= depositAmount;
+    checkingInput.val("");
+    savingsDisplay.html("$"+ savingsBalance);
+    savingsDisplay.css("background-color","red");
```
mattscilipoti, line 8: Think about the difference between assigning an attribute and setting a class here.

atm.js
```diff
@@ -1,21 +1,101 @@
...
+      savingsInput.val("");
+      savingsDisplay.html("$"+ savingsBalance);
+      savingsDisplay.css("background-color","red");
+    }
+}
+
+savingsDisplay.html("$"+ savingsBalance);
+checkingDisplay.html("$"+checkingBalance);
+
+$(".checking_account input[value=Deposit]").on("click",function(){
```
mattscilipoti, line 78: Be wary of using the text on the screen as a selector.  If you change the word in used in the button, you break your css.

atm.js
```diff
@@ -1,21 +1,101 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+function overDraft(){
```
mattscilipoti, line 2: +1 for extracting the behavior for an overdraft.

atm.js
```diff
@@ -1,21 +1,101 @@
...
+savingsDisplay.html("$"+ savingsBalance);
+checkingDisplay.html("$"+checkingBalance);
+
+$(".checking_account input[value=Deposit]").on("click",function(){
+    if (isNaN(this) == false){
+        checkingDeposit;
+    }
+    });
+$(".checking_account input[value=Withdraw]").on("click", function(){
+      if (isNaN(this) == false){
```
mattscilipoti, line 84: What is prompting us to verify that the item we clicked on is NotANumber?

mattscilipoti:
:+1: Looking good.  I left some comments inline.

That check for isNaN is throwing off the event handling.

PR #41, opened 2015-07-10
ATM

silviascript:
https://github.com/siltastic/atm/tree/silvia-solution

Comfort Level: 2/5 -> 3.5/5
Completeness: 3.5/5

I created functions to deposit and withdraw money from both accounts. They worked, but I have a syntax error that I'm trying to fix. The last function I'm still working on is the overdraft protection. I have a work in progress function (if and else) and adding a class commented out at the bottom of the page.

### DIFF
--- a/atm.css
+++ b/atm.css
@@ -2,43 +2,43 @@
 @import url(http://fonts.googleapis.com/css?family=Ruluko);
 
 body {
-  font-family: 'Ruluko', sans-serif;
-  margin: 0px auto;
-  text-align: center;
-  width: 700px;
+	font-family: 'Ruluko', sans-serif;
+	margin: 0px auto;
+	text-align: center;
+	width: 700px;
 }
 
 .header {
-  margin: 50px 0px;
+	margin: 50px 0px;
 }
 
 .title {
-  color: #F52F4F;
-  font-family: 'Oleo Script', cursive;
-  font-size: 50px;
+	color: #F52F4F;
+	font-family: 'Oleo Script', cursive;
+	font-size: 50px;
 }
 
 .account {
-  background-color: #6C9A74;
-  border: 1px solid #295A33;
-  border-radius: 15px;
-  display: inline-block;
-  margin: 10px;
-  padding: 10px;
-  width: 300px;
+	background-color: #6C9A74;
+	border: 1px solid #295A33;
+	border-radius: 15px;
+	display: inline-block;
+	margin: 10px;
+	padding: 10px;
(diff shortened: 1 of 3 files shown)


atm.css
```diff
@@ -2,43 +2,43 @@
 @import url(http://fonts.googleapis.com/css?family=Ruluko);
 
 body {
-  font-family: 'Ruluko', sans-serif;
```
adambray, line 5: It looks like your text editor is replacing soft tabs (two normal spaces) with hard tabs (a tab character). Please make sure Atom (or whatever your editor is), is configured to use soft tabs. Let me know if you need help with this.

atm.js
```diff
@@ -1,21 +1,89 @@
...
-function savingsDeposit(amount){
+	$("#addToChecking").on("click", function() {
 
-}
+		var depositMoneyToChecking = parseInt($("#checkingInput").val());
+		var checkingBalance = parseInt($("#checkingBalance").text().replace("$", "")); // Had lots of help here since innerHTML wasn't working for me.
+		var newCheckingBalance = "$" + (checkingBalance + depositMoneyToChecking);
+		
+		$("#checkingBalance").text(newCheckingBalance);
+			console.log("Added $" + depositMoneyToChecking + " to the checking account.");
```
adambray, line 23: make sure that your indentation is consistent (in general you're doing pretty well, but your console.log lines are indented too far).

atm.js
```diff
@@ -1,21 +1,89 @@
...
+	// User should be abe to deposit money to checking.
+	// User should be able to view current balance.
+	// User should be able to view updated balance.
 
-function savingsDeposit(amount){
+	$("#addToChecking").on("click", function() {
 
-}
+		var depositMoneyToChecking = parseInt($("#checkingInput").val());
+		var checkingBalance = parseInt($("#checkingBalance").text().replace("$", "")); // Had lots of help here since innerHTML wasn't working for me.
```
adambray, line 19: `innerHTML` is only for 'vanilla' DOM elements. If you're using jQuery, you need to use `text` as you did here.

adambray:
Good job overall! I don't have many other comments, since your variables are well-named, and overall your code is good.

Other thoughts:
1. It looks like you're on the right path for handling overdrafts. I strongly encourage you to finish this part of the exercise.
2. I noticed a lot of near-duplication in your code (the four lines to get the values, add / subtract, and update the display). Wonder if you could extract that into a method that could be re-used.

PR #42, opened 2015-07-10
Homework - ATM

malissaromero:
comfort level - 3
completeness - 5 (without bonus)

I included psuedo code at the end. I also attempted to try the bonus but left the code commented ouy because i could not get it to work.

i had a really hard time with user input. i kept putting the "enter an amount" into my running balance. Nick, the TA actually gave me the answer that it needs to be put into the user input variable.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,81 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready(function(){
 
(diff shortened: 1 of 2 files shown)


mattscilipoti:
:+1:  Looks like you are getting the hang of it.

It may help to suffix DOM related variables.  You did this for the buttons, but not the input or the balance (e.g. checkingAmountInput, checkingBalanceTag)

Thanks for leaving traces of your efforts.  It doesn't look like your overdraft protection code is removing the funds from the savings account, like your pseudo code indicates.

PR #43, opened 2015-07-10
finihed homework

Thesheik:
Confort level 3
completeness 3


mattscilipoti:
:+1:   

Watch your indentation.  It's much harder to read and reason about without it.

You have the functionality for the Checking deposit.  The withdraw is pretty much the same, but it removes money.  I would love to see you use the deposit steps and inform how to make the withdraw work.  It looks like you started on it, but you skipped steps.  Remember, trust each step before moving on.

PR #44, opened 2015-07-10
atm kinda

isabellabrookes:
completeness 3/5
comfort level 3/5

I'm actually really proud of what I accomplished so far.

The withdraw function works fine but it doesn't work when I put it inside the commented out if else statement.
I'm not sure if it's scope or if I'm just way off the mark there.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,75 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$("input").eq(0).attr("id", "cInBox"); //create id for checking account input box
+$("input").eq(1).attr("id", "cDeBut"); //create id for checking deposit button
+$("input").eq(2).attr("id", "cWiBut"); //create id for checking withdraw button
+$("input").eq(3).attr("id", "sInBox"); //create id for savings account input box
+$("input").eq(4).attr("id", "sDeBut"); //create id for checking deposit button
+$("input").eq(5).attr("id", "sWiBut"); //create id for checking withdraw button
+$(".balance").eq(0).attr("id", "cBalTot");//create id for checking balance total
+$(".balance").eq(1).attr("id", "sBalTot");//create id for savings balance total
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+// Removes the $ from the balance
+cBalTot.innerHTML.replace("$", "")
+// sum checking balance
+  var cBal = 0;
+  cBalTot.innerHTML.replace("$", "");
+$("#cDeBut").on("click", function(){
+  var depCBal = parseFloat($("#cInBox").val());
+  cBal = (depCBal + cBal);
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,75 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$("input").eq(0).attr("id", "cInBox"); //create id for checking account input box
+$("input").eq(1).attr("id", "cDeBut"); //create id for checking deposit button
+$("input").eq(2).attr("id", "cWiBut"); //create id for checking withdraw button
+$("input").eq(3).attr("id", "sInBox"); //create id for savings account input box
+$("input").eq(4).attr("id", "sDeBut"); //create id for checking deposit button
+$("input").eq(5).attr("id", "sWiBut"); //create id for checking withdraw button
+$(".balance").eq(0).attr("id", "cBalTot");//create id for checking balance total
+$(".balance").eq(1).attr("id", "sBalTot");//create id for savings balance total
```
adambray, line 8: I like that you're using IDs here!

That said, I'd suggest just adding these IDs to the HTML itself (unless we say otherwise, you're allowed to modify the HTML directly in an assignment).

Also, I'd suggest more descriptive ID names, such as `checking-deposit-btn`, `savings-input`, etc

atm.js
```diff
@@ -1,21 +1,75 @@
...
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+// Removes the $ from the balance
+cBalTot.innerHTML.replace("$", "")
+// sum checking balance
+  var cBal = 0;
+  cBalTot.innerHTML.replace("$", "");
```
adambray, line 14: make sure to keep your indentation consistent. You should only indent inside of `if`s, loops, and `functions` (basically anytime you start a new curly brace pair.

atm.js
```diff
@@ -1,21 +1,75 @@
...
+$("#sWiBut").on("click", function(){
+      var wdSBal = parseFloat($("#sInBox").val());
+      sBal = (sBal - wdSBal);
+      $("#sBalTot").text("$"+ sBal);
 
-}
 
-function checkingWithdrawal(amount){
+      // if withdraw is more than balance, cannot withdraw from savings
+/*    if ($("sInBox").val() <= sBal){
```
adambray, line 42: I think using variables to hold the amounts would make this code easier to read/write, e.g:

``` js
var checkingBalance = parseFloat($("#cInBox").val());
var savingsBalance   = parseFloat($("#sInBox").val());
var withdrawlAmount = parseFloat($("#sInBox").val());
var totalBalance = checkingBalance + savingsBalance;

if (withdrawlAmount > totalBalance) {
  // can't overdraft
} else {
  // can overdraft
}
```

adambray:
Good job with the functionality that you did implement, I like that you broke the problem down and implemented as many sub-problems as you could... 

My suggestion would be to focus on keeping your variable names meaningful and descriptive, as that may make it easier to think about the code that you're writing, and thus easier to solve the problem.

PR #45, opened 2015-07-10
Robert solution

robertbenjamin:
Comfort_level: 4
Completeness: 4


atm.js
```diff
@@ -1,21 +1,65 @@
...
+var sInput = $("#sInput");
 
-}
+$("#cDeposit").click(function (e){
+	e.preventDefault();
+	var deposit = parseFloat( $("#cInput").val() );
+	var total = parseInt(cBalance.text().replace("$", ""));
+	var newTotal = "$" + (total + deposit);
+	cBalance.html(newTotal);
+})
```
adambray, line 17: This all looks good!

atm.js
```diff
@@ -1,21 +1,65 @@
...
+	var sTotal = parseInt(sBalance.text().replace("$", ""));
+	var combineTotal = cTotal + sTotal;
+	if(combineTotal - withdraw >= 0) {
+		if(cTotal - withdraw >= 0) {
+			console.log("Withdraw from checking!");
+			newTotal = "$" + (cTotal - withdraw);
+			cBalance.html(newTotal);
+		} else {
+			console.log("Withdraw from savings, overdraft protection!!");
+			sBalance.html( sTotal - (withdraw - cTotal) );
```
adambray, line 32: I think you may need to add a '$' sign here?

adambray, line 32: Your logic for this section is really clean, but you may be able to improve the readability of the code by extracting one more variable, like:

``` js
var overdraftAmount = withdraw - cTotal;
```

atm.js
```diff
@@ -1,21 +1,65 @@
...
+			cBalance.html( "$0" );
+		}
+	} else {
+		console.log("Can't withdraw, bitch!");
+	}
+})
 
-function checkingWithdrawal(amount){
+$("#sDeposit").click(function (e){
```
adambray, line 40: This has a lot of duplication with the `cDeposit` section above. Any way you could extract this out into a general function that could take an input to determine whether to act on checking or savings?

adambray:
Overall, great job! 

One way you could try and take this to the next level (and this applies to project 1 as well) is to try and use objects to model the 'data' portion of your app, and then have event listeners use those objects  (and their methods) to try and keep your UI code as clean as possible.

PR #51, opened 2015-07-10
Adding working solution

nolds9:
comfort: 4
completeness: 4
Had a difficult time originally because I was bashing my head against a broken code base, but finally realized the problem and restarted from scratch on thurday and got the atm to function. Unfortunately did not have time to tackle the bonus or refactor my code into an object, but will work on doing that this weekend.


mattscilipoti:
:+1:   I'm glad you got to the root of the problem.  

As always, nice organization, formatting, and consistent naming.  I expect this will help when you start to refactor for the next functionality.  It also demonstrates an understanding of the problem.

PR #56, opened 2016-03-01
Vbajwa solution

vbajwa:
javascript is starting to feel really good here. Hitting my stride as far as code confidence goes. Reading eloquent javascript when i can, but behind in lectures. Still issues interacting with DOM!!! need some remediation. Talk to robin tomorrow. Will update this tomorrow as well when i have more time to work on it.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -20,37 +20,49 @@ var userCheckingInput = function(){
 };
 
 var userSavingsInput= function(){
-  var sAmount = parseInt($checkingAmount.val());
-  $checkingAmount.val("");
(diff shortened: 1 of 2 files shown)


RobertAKARobin:
You're getting it! To improve this more, your next step would be to do it so that you don't need to write `checking` and `savings` all over the place, or have two functions that basically do the same thing but for different accounts. Don't remember if you were in the review this morning, but feel free to watch the screencast and check out the code! https://github.com/ga-wdi-exercises/atm/tree/robin-inclass-solution

PR #57, opened 2016-03-01
Patricks solution

pkmoran:
comfort_level: 5
completeness: 5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,41 +1,3 @@
-// As a user, I want to deposit money into one of the bank accounts
-// As a user, I want to withdraw money from one of the bank accounts
-// Make sure the balance in an account can't go negative. If a user tries to withdraw more money than exists in the account, ignore the transaction.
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,91 @@
...
+});
+checking.withdrawButton.on("click", function() {
+  checkingWithdraw();
+  if (parseInt(checking.balance.html().replace("$", "")) === 0) {
+    checking.balance.addClass("zero");
+  }
+});
+savings.depositButton.on("click", function() {
+  savingsDeposit();
+  if (parseInt(savings.balance.html().replace("$", "")) > 0) {
```
RobertAKARobin, line 82: You could make a function for getting the numeric value of a particular element!

``` js
function getNum(element){
  return parseInt(element.html().replace("$", ""));
}
getNum(savings.balance);
```

PR #59, opened 2016-03-02
Sanjaya solution

ssatyal:
completeness: 5
comfort: 4
object-oriented comfort: 3

Worked with Jesse on refactoring the code into an OOP model, seems to be just above my head when we went over a few features to implement (i.e. things possible with jQuery, this is in the atm2.js file but not really looking for feedback on that at the moment - may revisit in the future).

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -90,6 +90,7 @@ $(':button').eq(3).on("click", function() {
   }
 });
 
+//these functions check the balances and if zero sets the class to zero
 function checkColorSave(){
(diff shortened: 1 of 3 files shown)


RobertAKARobin:
Looks like some pretty gnarly math going on here. I don't remember if you were in the ATM review this morning, but feel free to check the screencast and code: https://github.com/ga-wdi-exercises/atm/tree/robin-inclass-solution

PR #60, opened 2016-03-02
Ryan's Solution

rkrolick:
completeness: 5
comfort: 4
object-oriented comfort: 4

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,107 @@
+/*
 var checkingBalance = document.querySelector("div#checking_balance");
 checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+*/
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+window.onload = function(){
+  var checking = new Account(0);
+  var savings = new Account(0);
 
+
+  var checkingDepositBtn = $("input[value='Deposit']").eq(0);
+  var checkingWithdrawBtn = $("input[value='Withdraw']").eq(0);
+  var checkingDisplay = $(".balance").eq(0);
+  var checkingInput = $("input[type='text']").eq(0);
+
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,107 @@
+/*
 var checkingBalance = document.querySelector("div#checking_balance");
 checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+*/
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+window.onload = function(){
```
nolds9, line 6: Since you're using jquery, would recommend wrapping you're whole code into a `$(document).ready(function() { .. });` call

atm.js
```diff
@@ -1,21 +1,107 @@
...
+  var checking = new Account(0);
+  var savings = new Account(0);
 
-}
 
-function checkingWithdrawal(amount){
+  var checkingDepositBtn = $("input[value='Deposit']").eq(0);
+  var checkingWithdrawBtn = $("input[value='Withdraw']").eq(0);
+  var checkingDisplay = $(".balance").eq(0);
+  var checkingInput = $("input[type='text']").eq(0);
```
nolds9, line 15: Since there was not any constraints against modifying the `html`, would recommend adding id's to assist with selection, in favor of using `eq` to improve readability

atm.js
```diff
@@ -1,21 +1,107 @@
...
 
-}
+  var savingsDepositBtn = $("input[value='Deposit']").eq(1);
+  var savingsWithdrawBtn = $("input[value='Withdraw']").eq(1);
+  var savingsDisplay = $(".balance").eq(1);
+  var savingsInput = $("input[type='text']").eq(1);
```
nolds9, line 20: how could we modify our constructor function definition or modify the instances created to attach these variables as properties on the appropriate object?

nolds9:
:tada: Great job on this assignment. Nice work incorporating what we've learned about prototypes and constructors!

PR #62, opened 2016-03-02
Jeffreyeaton solution

jeffreyeaton76:
Completeness: 5
Comfort level: 4
I abandoned all pretense of writing DRY code in favor of getting 100% functionality. I should have deleted more of the given JS file (like maybe all of it) and taken an entirely different approach but by the time I realized it I had spent too much time on the assignment.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -2,30 +2,41 @@ $(document).ready(function() {
 
 var checkingBalance = $("#checkingBalance");
 var checkingInput = $("#checkingInput");
-var newTotal = 0;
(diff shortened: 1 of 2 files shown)


joe-gz:
:+1: Great job here Jeff! We talked about some things during our 1-1, so I don't have much by way of additional comments in line.  I know you were working on an object oriented version of this as well; that would be great! I think as you know, there's a lot of repetitive code with only "savings" and "deposit" changing.. definitely could call for some objects!

PR #66, opened 2016-03-03
Keith Prifte's ATM hw

khenry10:
completeness: 2
comfort: 3 (struggled for a while with the jquery, then starting refactoring into an object, and then I ran out of time)

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,10 +1,46 @@
+//Need to target the checking placeholder text box and then put an event listener on it.
+  //checkingAmountInput.prop("placeholder")
+// [x] Need to target the checking display amount
+  //checkingBalance2 = $(".balance").eq(0)
+//[x] Need to target the deposit checking button
+  //var checkingDepositButton = buttons.eq(0);
+// Need a function for the event listener to catpure the user input and print the amount on the screen
+
+atm = {
+  accounts: {
+    checking: 0,
+    savings: 0
+  }
+}
+
+var buttons = $("[type='button']");
+var checkingDepositButton = buttons.eq(0);
+
+var checkingAmountInput = $("[type='text']").eq(0)
+var checkingAmountInputText =  checkingAmountInput.prop("placeholder")
+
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,60 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+//Need to target the checking placeholder text box and then put an event listener on it.
+  //checkingAmountInput.prop("placeholder")
+// [x] Need to target the checking display amount
+  //checkingBalance2 = $(".balance").eq(0)
+//[x] Need to target the deposit checking button
+  //var checkingDepositButton = buttons.eq(0);
+// Need a function for the event listener to catpure the user input and print the amount on the screen
+
```
joe-gz, line 8: Love the notes here!

atm.js
```diff
@@ -1,21 +1,60 @@
...
+atm.buttons.checkingDeposit.on("click", checkingDeposit)
+atm.buttons.checkingWithdrawal.on("click", checkingWithdrawal)
+
+atm.buttons.savingsDeposit.on("click", savingsDeposit)
+atm.buttons.savingsWithdrawal.on("click", savingsWithdrawal)
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+function checkingDeposit() {
+  atm.accounts.checking = +atm.accounts.checking + +atm.display.checkingInputBox.val()
```
joe-gz, line 36: these extra "+" marks is a clever solution to this! But what if you just did something like:
`atm.accounts.checking += parseInt(atm.display.checkingInputBox.val())`

joe-gz:
:+1: Nice start on this Keith! Really like the approach you've taken. I don't have a lot by way of in line comments, but if you want to chat through what the next steps would be, I'm happy to sit down!

PR #67, opened 2016-03-03
Bipin solution

ghost:
complete 4
comfort 3.5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,36 +1,57 @@
-$(document).ready(function(){
-  // console.log(jQuery);
-  $("#depositCheckingButton").click(function(){
-    // console.log("clicked");
-    $("#checkingDepotext").val().append(".checkingBalance");
-  });
+//checking
+var checkingDepoButt = $("#checking_deposit")
+var checkingWithdrawButt = $("#checking_withdraw")
+var checkingInput = $("#checking_amount")
+var checkingBalance = $("#checking_balance")
+var checkingTotal = 0
+//savings
+var savingsDepoButton = $("#savings_deposit")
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,57 @@
...
+var checkingInput = $("#checking_amount")
+var checkingBalance = $("#checking_balance")
+var checkingTotal = 0
+//savings
+var savingsDepoButton = $("#savings_deposit")
+var savingsWithdrawButt = $("#savings_withdraw")
+var savingsInput = $("#savings_amount")
+var savingsBalance = $("#savings_balance")
+var savingsTotal = 0
+
```
amaseda, line 13: Hm, the two above chunks of code look pretty similar. Implementing an OOJS approach could help you condense this (and some of the code that appears later on in this file)...

Think about what an Account constructor function would look like. Passing in the DOM representation of an account (i.e., a `<div>` with the class `account`) as an argument to the constructor would allow you to target the buttons, inputs, etc. that belong to a given account without having to hardcore separate variables for your checking and savings accounts.

``` js
}
```

atm.js
```diff
@@ -1,21 +1,57 @@
...
+//withdraw amount from the savings account
+savingsWithdrawButt.click(function(){
+  //this is comparing the amount between balance and input
+  if (savingsBalance.html() < savingsInput.val() ) {
+    savingsBalance.addClass("zero");
+    alert("$$Ain't got no monay$$");
+  }
+  
+  else
```
amaseda, line 53: Missing some curly brackets here.

amaseda:
:+1: Let me know if you have any particular questions/comments about this assignment.

PR #71, opened 2016-03-03
John Blakeman solution

jblakeman:
Felt comfortable with this exercise.  Was able to get the objects and methods operational without much difficulty.  However, ran into an obstacle or two when I can time to throw listeners into the mix as it required passing `this` around.

Although I'm proud of myself for remaining pretty DRY here (save for one or two minor places that I didn't have time to fix) it got me into trouble when trying to add the listeners in a loop.  Will not do that again.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,89 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$("body").ready(function() {
+    function abs(num) {
+        if (num < 0) { return - num; }
+        return num;
+    }
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+    function Account(balance, input, deposit, withdraw) {
+        this.els = {
+            balance: balance,
+            withdraw: withdraw,
+            deposit: deposit,
+            input: input,
+        };
+        this.balance = 0;
+    }
+    Account.prototype.deposit = function(self) {
+        var amount = parseFloat(self.els.input.val());
+        if (amount > 0) {
+            self.balance += amount;
+            self.els.balance.html("$" + self.balance);
+        }
+    };
 
-}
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,86 @@
...
+        if (amount < 0) { 
+            console.log("Must withdraw a positive number");
+            return false;
+        }
+        this.protectOverdraft(amount, account, self);
+    };
+    var inputs = $(".account input[placeholder=\"enter an amount\"]");
+    var deposits = $(".account input[value=Deposit]");
+    var withdraws = $(".account input[value=Withdraw]");
+    var balances = $("div .balance");
```
joe-gz, line 68: Maybe consider moving these to the top? just a stylistic thing

atm.js
```diff
@@ -1,21 +1,86 @@
...
+    }
+    Accounts.prototype.drawRemainder = function(amount, account, self) {
+        if (!self) self = this;
+        remaining = self[account].balance - amount;
+        if (remaining < 0) { return abs(remaining); }
+        return 0;
+    };
+    Accounts.prototype.protectOverdraft = function(amount, account, self) {
+        var alternate = "checking";
+        if (account === "checking") { alternate = "savings"; }
```
joe-gz, line 36: clever!

atm.js
```diff
@@ -1,21 +1,86 @@
...
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$("body").ready(function() {
+    function abs(num) {
+        if (num < 0) { return - num; }
+        return num;
+    }
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+    function Account(balance, input, deposit, withdraw) {
```
joe-gz, line 7: Really like what you've done here with these constructor functions!

joe-gz:
:+1: Great job John! really clean work using constructor functions here. Let me know if you have any questions!

PR #73, opened 2016-03-03
Maureen assignment

maureenv:
I spent many hours on this assignment and I had a good start but it all went down hill from there. I couldn't get the numbers to add up and subtract correctly depending on whether I pushed the deposit or withdraw button. I will study the solution and try again.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,4 +1,65 @@
-var checkingBalance = document.querySelector("div#checking_balance");
+function ATM(){
+  this.els = {
+    checkingBalance: $(".balance").eq(0),
+    checkingInput: $("input").eq(0),
+    button: $(".checkingButton")
+  }
+}
+
+ATM.prototype.calculateTotal = function(input){
+  var input = input;
+  console.log(input);
+
+//console.log(input + "this is my input");
+  //for (var i =0; i<this.balance.length; i++){
+  //console.log("this is my balance " + total);
+//}
+  return input;
+}
+
+var checking = new ATM();
+
+checking.els.button.on("click", function(){
+  var buttonValue = ($(this).val())
+  //console.log("this is my button " + this.attr());
+  //var total=checking.calculateTotal ($(this).val());
+  var input = parseFloat(checking.els.checkingInput.val()); // if not a number is entered NaN is returned
+  var negativeInput = input * -1;
+
+  if (buttonValue==="Withdraw"){
+    var total=checking.calculateTotal (negativeInput);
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,4 +1,65 @@
-var checkingBalance = document.querySelector("div#checking_balance");
+function ATM(){
+  this.els = {
+    checkingBalance: $(".balance").eq(0),
```
mattscilipoti: Why convert this away from a jQuery object?  Why only convert two?

atm.js
```diff
@@ -1,4 +1,65 @@
...
+}
+
+ATM.prototype.calculateTotal = function(input){
+  var input = input;
+  console.log(input);
+
+//console.log(input + "this is my input");
+  //for (var i =0; i<this.balance.length; i++){
+  //console.log("this is my balance " + total);
+//}
```
mattscilipoti, line 16: I like to see instrumentation, for debugging.  However, this one confuses me.  I don't see anything that has a "balance" method and the loop doesn't use the "i" var that you are incrementing.

atm.js
```diff
@@ -1,4 +1,65 @@
-var checkingBalance = document.querySelector("div#checking_balance");
+function ATM(){
+  this.els = {
+    checkingBalance: $(".balance").eq(0),
+    checkingInput: $("input").eq(0),
+    button: $(".checkingButton")
```
mattscilipoti, line 5: These elements are unique, why not use the ID?  Then you can avoid the `eq()`.

atm.js
```diff
@@ -1,4 +1,65 @@
-var checkingBalance = document.querySelector("div#checking_balance");
+function ATM(){
+  this.els = {
```
mattscilipoti, line 2: :+1: I like this use of 'els'

atm.js
```diff
@@ -1,4 +1,65 @@
...
+  //var total=checking.calculateTotal ($(this));
+  var checkingBalance = checking.els.checkingBalance.html("$" + input); // changes the large number
+  checking.els.checkingInput.val(" ");
+})
+
+
+
+
+
+ /* JQuery
```
mattscilipoti, line 45: :+1:  Good idea.  Bring your notes to where you need them.

atm.js
mattscilipoti, line 5: I see why you did it for ".checkingButton".  Good answer to that problem.

mattscilipoti:
:+1: Good job.

PR #75, opened 2016-03-03
Brendan Magee solution - ATM homework

brendaneamon:
Comfort: 2.5
Completion: 3

I hit a wall when I was trying to implement overdraft protection and had to call it a day. I really wanted to implement the ATM as an object, but I struggled with getting started.  I'm not sure that the path I ended up taking is a good one, but I expect that more practice with these concepts will improve my understanding of them.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,22 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready(function(){
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
+
+
+var checkingDepositBtn = $("#checkingDepositBtn");
+
+var checkingDisplay = $("#checkingAccount > .balance");
+
+var checkingBalance = parseInt(checkingDisplay.html().replace("$" , ""));
+
+
+checkingDepositBtn.click(depositMoney);
+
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,63 @@
...
+
+  var checkingDepositBtn = $("#checkingDepositBtn");
+  var checkingWithdrawalBtn = $("#checkingWithdrawalBtn");
+  var checkingDisplay = $("#checkingAccount > .balance");
+  var checkingBalance = parseInt(checkingDisplay.html().replace("$" , ""));
+  var savingsDepositBtn = $("#savingsDepositBtn");
+  var savingsWithdrawalBtn = $("#savingsWithdrawalBtn");
+  var savingsDisplay = $("#savingsAccount > .balance");
+  var savingsBalance = parseInt(savingsDisplay.html().replace("$" , ""));
+
```
amaseda, line 11: I think the key to understanding why an OOJS approach is helpful here is visualizing what the input of the constructor function would be. You've probably already guessed that in an ideal world you wouldn't have to define separate variables and methods for your checking and savings account. Wouldn't it be swell if a constructor took care of that for us, regardless of whether we're working with a checking, savings or any other account?

The below is some pseudo-code that might help you get started. Happy to talk about this more during a 1:1!

``` js
}
```

amaseda:
:+1: Good job with this. I think starting to implement the OOJS approach I described in my code comment will help a lot with refactoring. Let me know if you have any additional questions/comments.

PR #76, opened 2016-03-03
AndrewS-M solution

arsm800:
comfort: 3.75
completeness: 4

Based on our lesson from today, I know there are ways to refactor the code so that I have fewer functions.  The deposit and withdrawal functions are prime candidates to streamline.  Any suggestions on the best way to do it?

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,9 +1,24 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingBalance = $(".balance").eq(0);
+// checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var savingsBalance = $(".balance").eq(1);
 
+//var checkingDeposit = function(amount){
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,21 +1,99 @@
...
+  checkingBalance: 0,
+  savingsBalance: 0
+};
 
+var buttons = {
+  chkDepButton: $("input").eq(1),
+  chkWitButton: $("input").eq(2),
+  savDepButton: $("input").eq(4),
+  savWitButton: $("input").eq(5)
+};
```
amaseda, line 14: Can you think of how creating an `Account` constructor would help you refactor `var buttons`? Imagine if you passed in a DOM element to the constructor - how could you use that to reference its checking and savings buttons?

``` js
}
```

amaseda:
Echoing my previous comment, but I think the best way to refactor this is to implement an OOJS approach.

As your code stands, you've manually created a variable to represent each element on the page. Imagine if you were managing 10 different accounts with this app. That means, for example, that `buttons` would contain 20 variables!

What makes the OOJS approach nice here is that by passing in an element to a constructor, you can use `this` to reference any buttons, inputs, etc. that belong to a particular account. So in the case of buttons, what could have been a list of 20 variables representing 20 different buttons can now be shrunk down to 2. Happy to talk through this in person!

amaseda:
:+1: Good job with this.

PR #77, opened 2016-03-03
ATM HW submission

Ceisen9:
Initial commit, accidentally did not fork the atm repo before I cloned it down. Had to paste in code from local repo to push back to the remote repo.
Code is still very WET. Needs significant DRYing on my part, but it's working up to snuff.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,101 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
+function Account(value){
+  this.entries = [];
+  this.els = {
+    input: $("input[type=text]").eq(value),
+    deposit: $('input[type=button][value="Deposit"]').eq(value),
+    withdraw: $('input[type=button][value="Withdraw"]').eq(value),
+    total: $(".balance").eq(value)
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,101 @@
...
+  var total = 0;
+  for( var i = 0; i < this.entries.length; i++ ){
+    total += this.entries[i];
+  }
+  return total;
+};
 
-function savingsDeposit(amount){
+var checking = new Account(0);
+var savings = new Account(1);
```
jshawl, line 20: clever! you could also pass in a selector to specify which account you want, e.g:

``` js
var checking = $(".account").eq(0);
new Account( checking )
```

atm.js
```diff
@@ -1,21 +1,101 @@
...
 
+function balanceChecker() {
```
jshawl, line 22: could add this to a prototype method: `Account.prototype.getBalance()`

atm.js
```diff
@@ -1,21 +1,101 @@
...
+  var currentSavingsBalanceString = savings.els.total.html();
+  var currentSavingsBalanceNum = Number(currentSavingsBalanceString.replace(/[^0-9\.]+/g,""));
```
jshawl, line 24: nice use of regex!

jshawl:
:+1: Excellent job on this assignment! Let me know if you'd like to go through the DRYing process together. My general process is to look for lines of code where everything is the same but one thing, then create a function and pass in that one thing that's changing. keep up the great work!

PR #78, opened 2016-03-03
Fatai solution

yinkaagiri:
Completeness 1
Comfort level 3
Timebox: 3 hrs . I tried to use vanilla JS because the starter code is in Vanilla JS. I will refactor and continue with Jquery

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -7,7 +7,7 @@ var checkingDeposit = function(amount){
 }
 
 function checkingWithdrawal(amount){
-
+ return
 }
 
 function savingsDeposit(amount){
--- a/atm.js
+++ b/atm.js
@@ -1,13 +1,17 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checking_balance = document.querySelector("#checking_balance");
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,9 +1,13 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
+var checking_balance = document.querySelector("#checking_balance");
+checkingBalance = parseInt( checking_balance.innerHTML.replace("$", "") );
```
jshawl, line 2: instead of getting the balance value from the DOM, you can create a variable to hold the balance, and update the `innerHTML` everytime the balance changes

atm.js
```diff
@@ -1,9 +1,13 @@
...
+savings_amount = querySelectorall("savings_deposit").value;
+checking_amount = querySelectorall("savings_deposit").value;
```
jshawl, line 4: This will cause `Uncaught ReferenceError: querySelectorall is not defined`. use `document.querySelectorAll()` or `$("savings_deposit")` note: `.` in selector string

jshawl:
you're off to a good start! Let's find some time this week to work through this one together.

PR #80, opened 2016-03-03
Adrian Wyatt "Solution"

Wyatt23:
Update: 
comfort level: 3.5
completeness: 2.5
what I needed to realize was that $(document).ready, which was originally enveloping my whole page would cause the issue with my error messages in the console. I left my dead code in there so you could see my work. A lot was explained during home work review that could have helped me out  - the general idea I had, I would have played with it, but my real road block was that I spent too much time trying to add classes using jQuery and struggling to get them to let me reference them in the same document - I thought we were not supposed to touch the HTML. Dern it. ;) 

comfort level: 2
completeness:2 

So... I was really excited to make these accounts into constructor functions and send the slight differences up as prototypes to new accounts - but... I was halted...for HOURS because I cannot figure out how to make a click function apply to classes I assigned from the JS file. 

I can use jQuery to assign classes and id's etc. but I can't make a single input into a storable thing when I use them. Please help me so I can continue the fun part - seeing if I can make these accounts into constructors. (also, there's a bit of code for a whole lot of no action on the HTML page. Sry. I tried) 

:)

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,6 +1,29 @@
-$("input").eq(0).addClass("checking_input");
-$("input").eq(3).addClass("savings_input");
+var checkingUserInput = ($("body > div:nth-child(2) > input[type="text"]:nth-child(3)").val());
+var checkingDepositButton = ($("body > div:nth-child(2) > input[type="button"]:nth-child(4)").val());
+var checkingWithdrawlButton = ($("body > div:nth-child(2) > input[type="button"]:nth-child(5)").val());
+var checkingBalance = ($("body > div:nth-child(2) > div").val());
 
+// $("input").eq(0).addClass("checking_userInput");
+// $("input").eq(1).addClass("checking_deposit_button")
+// $("input").eq(2).addClass("checking_withdrawal_button")
+// $("input").eq(3).addClass("savings_userInput");
+// $("input").eq(4).addClass("checking_deposit_button")
+// $("input").eq(5).addClass("checking_withdrawal_button")
+
+var checkingTotal = 0
+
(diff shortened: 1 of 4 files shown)


atm.js
```diff
@@ -1,21 +1,71 @@
...
+  }
+  account.elements.deposit.click(function(){
+    var input = parseInt(account.elements.input.val());
+    account.balance += input;
+    account.elements.balance.html("$" + account.balance);
+  })
+  account.elements.withdraw.click(function(){
+    var input = parseInt(account.elements.input.val());
+    account.balance -= input;
+    account.elements.balance.html("$" + account.balance);
```
amaseda, line 24: Next step: how can you prevent withdraw from being executed when the amount-to-be-withdrawn is greater than the current account balance?
- What two values would you need to compare? How can you reference them?
- Assuming withdraw > deposit, what Javascript features could you use to prevent any modification to the balance (both the data and its representation in the DOM)?

atm.js
```diff
@@ -1,21 +1,71 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready(function(){
```
amaseda, line 1: I believe you were referencing not being able to test values in the console in your PR description. Important to remember that by placing things in `$(document).ready(function(){...})` we are removing them from the global scope.

amaseda:
:+1: Good job with this. Glad the homework review helped. Let me know if you have any questions about how you can take this assignment further.

PR #82, opened 2016-03-03
Yun solution

hh747hh:
still working on it...


RobertAKARobin:
Looking pretty good so far! Totally a minor note, but I'd encourage you to write your HTML attributes like this:

``` html
<div class="hello"></div>
```

...instead of like this:

``` html
<div class = "hello"></div>
```

PR #86, opened 2016-03-03
Anissa solution

iamanissa:
Comfort score: 4 out of 5
Completeness score: 4 out of 5
Any comments: I was working on adding draft protection, I deleted what I wrote b/c I wanted my code to work upon turning it in, however, if someone would be willing to look at my repo I think 1 commit prior to this, pull request I would like feedback on if I was going in the right direction. 
Right now my code just ignores the users request to take out more money then they have in one bank account, when I added my code, it would let them take it out if they both covered it but then other issues happened when trying to run tests that worked before I added the new code.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -4,21 +4,42 @@ $(document).ready(function(){
     //get innerhtml from balance div
 
 
-
+    //gets the jquery deposit button element
     var checkingDepositBtn = $("#checkingDepositBtn");
 
-    var checkingDisplay = $("#checkingAccount >.balance")
+    var checkingWithdrawBtn = $("#checkingWithdrawBtn");
 
+    //getting the jquery checking balance element.
+    var checkingDisplay = $("#checkingAccount >.balance");
+
+    //getting the current value of the balance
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,91 @@
...
+    var savingsDepositBtn = $("#savingsDepositBtn");
 
-}
+    var savingsWithdrawBtn = $("#savingsWithdrawBtn");
 
-function savingsWithdrawal(amount){
+    var savingsDisplay = $("#savingsAccount >.balance");
 
-}
+    var savingsAmount = parseInt(savingsDisplay.html().replace("$", ""));
```
amaseda, line 23: How could you use OOJS to create a constructor function that would save you the trouble of having to create variables that store buttons, inputs, etc. for each account.

``` js
}
```

amaseda:
Let's plan on going over your overdraft code from your earlier commit during our 1:1.

amaseda:
:+1: Great job with this assignment. Think about my comment re: OOJS. It could help you do some major refactoring!

PR #87, opened 2016-03-03
Rob's solution

Robizzy27:


atm.js
```diff
@@ -16,6 +16,35 @@ function savingsDeposit(amount){
 
 function savingsWithdrawal(amount){
 
+}*/
+
+// an eventListerner for each button, each one a "click"
+function checkingBalance(){
+  this.entries = [];
+  this.els = {
+    form: $("#entry"),
```
nolds9, line 25: do we have any elements with `id`s that match these selected?

atm.js
```diff
@@ -16,6 +16,35 @@ function savingsDeposit(amount){
...
+checkingDeposit.els.form.on("submit", function(event){
+  event.preventDefault();
+  var newDeposit = parseFloat(checkingDeposit.els.input.val())
+  console.log(newDeposit)
+  checkingDeposit.els.input.val("")
+  checkingDeposit.entries.push(newDeposit);
+  var newBalance = checkingDeposit.calculateTotal()
+    checkingDeposit.els.total.html("$" + newBalance)
+
+})
```
nolds9, line 50: Looks like a good start logic wise, but would recommend breaking down each step as it applies to this assignment. 

What are the element we care about? 
How could we select and store those elements so as to be able to access those values later?
What functionality is shared across both accounts?

nolds9:
Please come back to this when you have to time and build off of your work.

PR #88, opened 2016-03-03
Winnies solution

winishahines:
comfort : 2
completeness: 0

Wasn't sure where to begin with this one. I looked at the cash register example from class but was confused.


RobertAKARobin:
I wish you'd let me know earlier! :(

PR #89, opened 2016-03-03
charleshwang solution

ghost:
comfort: 3.5/5

completeness: 3.5/5

I approached this assignment the same way we did the cash register assignment in class using constructor functions and prototypes (not sure if this was the best approach). I was able to get each account functioning but got stuck "transferring" funds from savings to checking. If the checking withdrawal amount was greater than checking balance, I was able to set the checking balance to 0, but unable to pull the remaining withdrawal amount from savings. I _think_ the issue was trying to reference a variable that was only defined in my savings constructor/prototype functions? Suggestions?

Will continue to work on this, if time permits...

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -42,12 +42,17 @@ checking.prototype.calculateBal = function() {
   for (var i = 0; i < this.entries.length; i++) {
     total += this.entries[i];
   }
+  // changes color to red if balance hits zero
   if (total === 0) {
     $("#checkingAcct").addClass("zero");
     return total;
   } else if (total >= 0) {
     $("#checkingAcct").removeClass("zero");
     return total;
+  // if userInput > checking balance, set to 0; pull (userInput - checking balance) from savings
+  } else if (total < 0) {
+    $("#checkingAcct").addClass("zero");
+    return 0;
   }
 };
 
@@ -77,12 +82,8 @@ myChecking.els.checkingWithdrawal.click( function(event) {
   // get user input
   var userInput = parseInt(myChecking.els.input.val());
   console.log(userInput);
-  // turn user iput to a negative value
-  if (userInput > 0) {
-    userInput = userInput * -1;
-  }
-  // add the inputs to the array of entries
-  myChecking.entries.push(userInput);
+  // add the negative value of inputs to the array of entries
+  myChecking.entries.push(-(userInput));
   // displays user input in checking box
   myChecking.els.value.html(userInput);
   // calculate balance
@@ -111,6 +112,7 @@ savings.prototype.calculateBal = function() {
   for (var i = 0; i < this.entries.length; i++) {
     total += this.entries[i];
   }
+  // changes color to red if balance hits zero
   if (total === 0) {
     $("#savingsAcct").addClass("zero");
     return total;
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,21 +1,192 @@
...
+//
+// }
+//
+// // an eventListerner for each button, each one a "click"
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+$(document).ready(function() {
 
+function checking() {
```
beckybeauchamp1, line 25: just remember to capitalize constructor functions :) 
`function Checking(){}`

atm.js
```diff
@@ -1,21 +1,192 @@
...
+  // displays user input in checking box
+  myChecking.els.value.html(userInput);
+  // calculate balance
+  var newBal = myChecking.calculateBal();
+  // displays new balance
+  myChecking.els.value.html("$" + newBal);
+});
 
+function savings() {
+  $("#savingsAcct").addClass("zero");
```
beckybeauchamp1, line 96: I think it's great how you used constructor functions, but i think you could have used one constructor function for both savings and checkings. So instead of two, just one maybe called account constructor:

```

function Account(type){
  this.name = type;
  this.entries = [];
  this.els = {
    value: $("#" + type + "Acct"),
    input: $("#" + type + "Amt"),
    deposit: $("#" + type + "Deposit"),
    withdrawal: $("#" + type + "Withdrawal"),
  };
}

Account.prototype.calculateBal = function() {
  var total = 0;
  for (var i = 0; i < this.entries.length; i++) {
    total += this.entries[i];
  }
  if (total === 0) {
    $("#" + this.name + "Acct").addClass("zero");
    return total;
  } else if (total >= 0) {
    $("#" + this.name + "Acct").removeClass("zero");
    return total;
  } else if (total < 0) {
    $("#" + this.name +"Acct").addClass("zero");
    return 0;
  }
};

function makeAccounts(){
  var savings = new Account("savings");
  var checking = new Account("checking");
}

makeAccounts();
```

atm.js
```diff
@@ -1,21 +1,192 @@
...
+    // savings deposit button
+    savingsDeposit: $("#savingsDeposit"),
+    // savings withdraw button
+    savingsWithdrawal: $("#savingsWithdrawal"),
+  };
 }
 
-// an eventListerner for each button, each one a "click"
\ No newline at end of file
+savings.prototype.calculateBal = function() {
```
beckybeauchamp1, line 110: Look at my suggestion above, but i think you could refactor and condense a lot of this but putting a calculateBal method on one Account constructor prototype instead of having two separate ones

atm.js
```diff
@@ -1,21 +1,192 @@
...
+  } else if (total >= 0) {
+    $("#savingsAcct").removeClass("zero");
+    return total;
+  }
+};
+
+//creating a new savings function based on savings constructor function
+var mySavings = new savings();
+
+// what happens when savings deposit button is clicked
```
beckybeauchamp1, line 128: Awesome job commenting your code! This is so great to see and so helpful :+1:

beckybeauchamp1:
Hi Charles,

Good job on your ATM homework! I provided additional feedback inline. To answer your question, it's totally fine using an OOP/constructor function approach, theres not a "right" way to solve this assignment. I think refactoring a little bit of your code would help you to solve the last problem with withdrawling from savings. 

Let me know if you have any more questions, happy to sit down with you further and work through this together! Great job :+1:

PR #91, opened 2016-03-03
Toni langley solution

tvcrawley:
Complete: 3.5
Comfort: 4

Comment: The started code was more confusing than helpful.


RobertAKARobin:
Thanks for the feedback!!!

PR #92, opened 2016-03-04
only works when I call checkingNewBalance = parseInt(.val()) in console

Crunchwrap78:
pull requests for atm and geometry didn't go through

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,21 +1,63 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checking =0;
+var savings = 0;
+var $checking_amount = $("#checking_amount");
+var $savings_amount = $("#savings_amount");
+var $checking_balance = $(".checking_balance");
+var $saving_balance = $(".saving_balance");
+$checking_amount.val(" ")
+savingsNewBalance = parseInt($($saving_balance.val())
 
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
+var checkingDeposit = function(){
+  checkingNewBalance = parseInt($checking_amount.val())
+  checking += checkingNewBalance;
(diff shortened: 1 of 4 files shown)


atm.js
```diff
@@ -1,21 +1,63 @@
...
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
+var checking =0;
+var savings = 0;
+var $checking_amount = $("#checking_amount");
+var $savings_amount = $("#savings_amount");
+var $checking_balance = $(".checking_balance");
+var $saving_balance = $(".saving_balance");
+$checking_amount.val(" ")
+savingsNewBalance = parseInt($($saving_balance.val())
```
jshawl, line 8: this line is causing a syntax error, need one more paren on the right

atm.js
```diff
@@ -1,21 +1,63 @@
-var checkingBalance = document.querySelector("div#checking_balance");
-checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
-var checkingDeposit = function(amount){
-  checkingBalance = parseInt( checkingBalance.innerHTML.replace("$", "") );
-
+var checking =0;
+var savings = 0;
+var $checking_amount = $("#checking_amount");
```
jshawl, line 3: oo nice naming convention

atm.js
```diff
@@ -1,21 +1,63 @@
...
+  savingsNewBalance = parseInt($($saving_balance.val())
+  savings -= savingsBalance;
+  $savings_balance.html("$"+savings)
+  if(savings === 0){
+    $savings_amount.addClass("zero")
+  }
+  else {
+    $savings_amount.removeClass("zero")
+  }
 }
```
jshawl, line 57: both the savingsWithdrawal and checkingWithdrawal have similar functionality, save for the value they are reading and where they are outputting the new balance. Can you think of a way to combine these two functions into one, using arguments to distinguish them?

jshawl:
:+1: Let me know if you'd like to work through the DRYing together.

PR #98, opened 2016-06-02
Lab complete - Sean Tay

SeanTay:
comfort level: 4/5

completeness: 5/5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,4 @@
 $('#checking input.deposit').on('click', function(){
-  console.log('money')
+  var depositAmt = $('#checking input.input').val();
+  console.log(depositAmt);
 })
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,4 @@
 $(document).ready(function(){
 // Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,134 @@
...
+          //add and remove class based on the balance
+          if (ScurrentBalance > 0){
+            $('#savings div.balance').removeClass('zero');
+          } else {
+            $('#savings div.balance').addClass('zero')
+          }
+
+          //display the currentBalance
+          $('#savings div.balance').html(ScurrentBalance);
+        }
```
RobertAKARobin, line 64: Kinda weird indentation!

atm.js
```diff
@@ -1,3 +1,134 @@
...
+
+
+})
+
+
+//checking deposit button function
+$('#checking input.deposit').on('click', function(){
```
RobertAKARobin, line 71: Notice that your checking and savings accounts do deposits and withdrawals the exact same way. Having a deposit function for checking and one for savings isn't too bad when you just have 2 accounts, but what if you had 200 accounts...?

RobertAKARobin:
✅ 👍

PR #99, opened 2016-06-02
My atm

AT89:
5/5 completeness
4/5 comfort level
 Starting the program was tough, I did the pseudocode but I did not know where to really begin. I began with the checking, got the deposit to work and then the withdraw. copied it and changed the tags for savings. Then I ran into a whole lot of bugs to get the overdraft to run. But I am glad that I was able to finish this on time (without staying after), so there was definitely enough time to get this lab done; if it took more time than the class time, I would be pretty uncomfortable.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -8,10 +8,11 @@ $(document).ready(function(){
     //deposit
     $("#checking .deposit").on("click", function(){
         //amount in box, click deposit-adds val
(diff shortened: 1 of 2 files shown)


RobertAKARobin:
Good job on tackling one functionality at a time! ✅

PR #104, opened 2016-06-02
Julia solution

jschultz21:
comfort: 3.7
complete: 5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -16,27 +16,42 @@ var checkingInput = $("#checking .input").val();
 
 $("#checking .deposit").on("click", function(){
 
-  console.log(checkingInput);
+  var checkingInput = $("#checking .input").val();
 
+    if (checkingInput <=0) {
+      alert("Please enter an amount greater than 0");
+      }
+
+    else {
+  // console.log(checkingInput);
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,148 @@
...
+  }
+
+  function getAccountColorB() {
+    if (savingsBalance==0) {
+      $("#savings .balance").addClass("zero");}
+
+      else {
+        $("#savings .balance").removeClass("zero");
+      }
+    }
```
RobertAKARobin, line 28: You've got some weird curly-bracing going on here. Keep it consistent.

Also, if you find yourself naming things like `itemA`, `itemB`, `itemC` and so forth, you can almost definitely extract those into a function to stay DRY. For example:

``` js
var balances = {
    checking: 0,
    savings: 0
}

function getAccountColor(acctName) {
    if (balances[acctName]==0) {
        $("#" + acctName + " .balance").addClass("zero");
    }else{
        $("#" + acctName + " .balance").removeClass("zero");
    }
}

getAccountColor("checking");
getAccountColor("savings");
```

RobertAKARobin:
✅

PR #105, opened 2016-06-03
Gandy atm exercise

gandyland:
completeness: 2
comfort: 2

This is not the code I ended with but I apparently didn't commit correctly and guess I lost some of it. Too frustrated to try and do fix it. Going to review git and jquery a lot this weekend.


RobertAKARobin:
This code looks like a good start, but it doesn't run due to syntax errors.

✅

PR #110, opened 2016-06-03
ATM Lab Courtney Seitz

Courtney-Seitz:
Comfort level - 3
Completeness - 3

I had to stop working on the overdraft protection bit after a while. I still have questions on that once I attempt it more. Turning in for now.


RobertAKARobin:
This appears to be taken very closely from the solution. That's fine, but be careful that it doesn't turn into a habit!

Please Slack more. I'll start nagging you if you don't.

✅

PR #115, opened 2016-06-03
i can and will finish this

BlvckYuppie:
didnt get home until 11 last night, was too knackered to write the savings side, but i'll finish it @ lunch today.


atm.js
```diff
@@ -1,3 +1,37 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+//These are wrong, should probably be parseInt's
```
amaseda, line 2: Are you referring to your selectors in `left`? They look fine to me. It's useful to define references to these DOM objects.

index.html
```diff
@@ -27,5 +27,7 @@
       <input class="deposit" type="button" value="Deposit" />
       <input class="withdraw" type="button" value="Withdraw" />
     </div>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.2.2/jquery.min.js"></script>
+    <script src="atm.js"></script>
```
amaseda, line 31: You link to `atm.js` twice in this file. This one is correct -- the one in `<head>` is being loaded prior to jQuery, which is why you might be seeing `Uncaught ReferenceError: $ is not defined` in your browser console.

amaseda:
Good job getting checking to work. I think you can easily apply the same to savings. Let me know if you have questions/comments about this or want to go through setting up overdraft.

PR #118, opened 2016-06-06
Ct gardner solution

gardner589:


atm.js
```diff
@@ -1,3 +1,75 @@
 $(document).ready(function(){
 // Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+var savings = {
```
amaseda, line 3: :+1: Organizing each account into an object -- nice! When we get into OOJS, you'll be able to refactor this even further so that `checking` and `savings` (and as many other accounts as you want) can be generating using a single object definition (or "constructor").

atm.js
```diff
@@ -1,3 +1,75 @@
...
+    if ( (savings.balVal - parseInt(savings.putted.val()) ) >= 0){
+      savings.balVal = savings.balVal - parseInt(savings.putted.val());
+      savings.balance.html("$" + savings.balVal.toString());
+      savings.putted.val("");
+
+      if(savings.balVal == 0){
+        checking.balance.addClass("zero");
+      }else{
+        checking.balance.removeClass("zero")
+      }
```
amaseda, line 16: You could shorten this by using `.toggleClass`. 

http://api.jquery.com/toggleclass/

atm.js
```diff
@@ -1,3 +1,75 @@
...
+    if( ( checking.balVal - parseInt(checking.putted.val()) ) >= 0){
+      checking.balVal = checking.balVal - parseInt(checking.putted.val());
+      checking.balance.html("$" +  checking.balVal.toString());
+
+    }else if( (checking.balVal + savings.balVal) > parseInt(checking.putted.val()) ){
+       var over = parseInt(checking.putted.val()) - checking.balVal;
+       savings.balVal = savings.balVal - over;
+       checking.balVal = 0;
+       savings.balance.html("$" + savings.balVal.toString());
+       checking.balance.html("$" + checking.balVal.toString());
```
amaseda, line 48: You have a few lines here that look very similar. Perhaps this means you could replace them with a function...

amaseda:
:+1: Great job with this.

PR #123, opened 2016-07-07
Megan's Completed ATM

megancoyle:
Comfort score: 4
Completeness: 5

Definitely felt better about this one than I did with the cash register, but that's probably because I was trying to do too many things with the last assignment. My main concern here is how to refactor this - I definitely repeated myself quite a bit, but at least it's working the way it should! Let me know if you see any weird habits in the way I'm structuring this, and any obvious ways to make this cleaner. Thanks!

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,5 +1,27 @@
 $(document).ready(function(){
 
+// define savings account
+// define checking account
 
+// define value of user input
+  //only accept numeric value
+
+// on click of deposit, user input added to total value of specific account
+   // if amount is > 0, remove zero class
+   // clear input field after amount is added
+
+// on click of withdraw, user input subtracted from total value of specific account
+  // if total amount = 0, apply zero class
+  // if amount to withdraw > total funds, ignore transaction
+  // clear input field after amount is withdrawn
+
+
+//Bonus
+
+// if total to withdraw is equal to total in both accounts
+// have current account total = 0
+// subtract remaining amount from other account
+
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,121 @@
...
+
+// starting balance with .zero
+$balance.addClass("zero");
+
+// on click of deposit or withdraw, user adds/subtracts balance
+$depositBtn.on("click", depositSomething);
+$withdrawBtn.on("click", withdrawSomething);
+
+
+function depositSomething(e) {
```
andrewsunglaekim, line 26: maybe leave off Something for variable name, it's a small thing, but something isn't really descriptive and deposit is enough semantics to clarify what the functions trying to do.

atm.js
```diff
@@ -1,3 +1,121 @@
...
+  e.preventDefault();
+  // check to see if ID of parent div is checking or savings
+  var $accountId = $(this).parent().attr("id");
```
andrewsunglaekim, line 29: really like that you don't have a `depositChecking()` and a `depositSavings()` function an instead call one and check the id. noice.

andrewsunglaekim:
Really great work megan! Although Taylor didn't get to all the same functionalities you did. He managed to abstract all of his functionalities into just one event listener. Also I would check out John from your squads solution as well. He had a really clever solution for abstraction as well. You three should pair and combine all of this great code! 👍

PR #124, opened 2016-07-07
Pamelas hw

pamelaolney:
Comfort: 2
Completeness: 5

I would like to be able to come up with these solutions on my own without much assistance.  My classmates walked me through every piece.  I plan  on really practicing these things over the weekend so I can be more confident of my coding skills.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,141 @@
 $(document).ready(function(){
 // Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
-});
+
+//change to red when balance is at 0
+$("#checking > .balance").addClass("zero");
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,141 @@
...
+            console.log("nuh uh")
+          } else {
+            $("#savings > .balance").html("$"+ actualSavingsWithdrawTotal.toFixed(2));
+
+        }
+
+        //clearfield checking input
+        $("#savings > .input").val("")
+      }
+
```
superbuggy, line 139: The only thing missing is a conditional checking to see if the input is blank before we make a deposit or a withdrawal. That and pseudo code, but other than that you nailed this lab!

PR #125, opened 2016-07-07
Finished Atm Project

ghost:
Comfort: 5
Completeness: 5

This was awesome! A lot of the cash register really helped with the bank! I'm really pumped to do the project for next week after this. Is there a way I can slim down this codebase any more?

Thanks!

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,4 @@
+//adds zero class
 function addThis(selector){
     $(selector).addClass(
         function(index,currentClass){
@@ -9,7 +10,7 @@ function addThis(selector){
         }
     )
 }
-
+//removes zero class
 function removeThis(selector){
     $(selector).removeClass(
         function(index,currentClass){
@@ -21,11 +22,11 @@ function removeThis(selector){
         }
     )
 }
+//Doc is loaded and primed
 $(document).ready(function() {
     //implements the zero class on start
     addThis(".balance");
-
-    //prevent enter from reloading page
+    //prevent enter from reloading page with form wrapper
     $('form').submit(function(e) {
         e.preventDefault();
     });
@@ -92,6 +93,8 @@ $(document).ready(function() {
         if(!isNaN(withdrawAmount) && difference >= 0){
             $('.balance:eq(1)').html('$' + (currentBalance - parseFloat(withdrawAmount)).toFixed(2));
             $('.entry:eq(1)')[0].reset();
+        } else {
+            alert('No Funds In Savings!')
         }
     });
 });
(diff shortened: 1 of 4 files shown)


atm.js
```diff
@@ -1,3 +1,98 @@
...
+function addRedBalance(selector){
+    if(parseFloat($(selector).html().split('$')[1])==0){
+        $(selector).addClass("zero")
+    }
+}
+//removes zero class
+function removeRedBalance(selector){
+    if(parseFloat($(selector).html().split('$')[1])!==0){
+        $(selector).removeClass('zero')
+    }
```
nayana487, line 11: hmmm....why two document ready functions? Since your making function declarations here, they can be anywhere in the code and still be called elsewhere because of hoisting.

atm.js
```diff
@@ -1,3 +1,98 @@
...
+    });
+
+    //deposit function checking
+    $(".deposit:first").on('click', function(event) {
+        event.preventDefault();
+        var depositAmount = parseFloat($('.input:first').val()).toFixed(2)
+        if(!isNaN(depositAmount)){
+            var currentBalance = parseFloat($('.balance:first').html().split('$')[1])
+            $('.balance:first').html('$' + (parseFloat(depositAmount)+currentBalance).toFixed(2));
+            addRedBalance(".balance:first");
```
nayana487, line 29: So the way you're choosing your selectors here isn't wrong, but I'd steer on the side of making it as simple and readable as possible. something like $(#checking .balance) is easy for other developers to read..the way you have it, you gotta think about a little. It's cool to learn how to use it, but in actuality that's one most of us would have to pause at.

atm.js
```diff
@@ -1,3 +1,98 @@
...
+            //overdraft protection logic
+        } else if (!isNaN(withdrawAmount) && difference < 0 && saveBalance > (difference*-1)) {
+                var protectionAmount = ((currentBalance - withdrawAmount) * -1).toFixed(2)
+                var savingsAmount = parseFloat($('.balance:eq(1)').html().split('$')[1])
+                $('.balance:eq(1)').html('$'+ (savingsAmount - parseFloat(protectionAmount)).toFixed(2))
+                $('.balance:first').html('$0.00')
+                addRedBalance(".balance:first");
+                addRedBalance(".balance:eq(1)");
+                alert('Careful! Overdraft Protection Used! $' + protectionAmount + ' removed from savings')
+                $('.entry:first')[0].reset();
```
nayana487, line 56: Nice job with the bonus here! The math is always the toughest part.

nayana487:
Your code here is pretty tight,  but one thought is that you probably didn't need to create separate functions for adding and removing the zero class. In this case, it's probably less to write out just applying the methods.

PR #128, opened 2016-07-07
Taylor's lab (7/7)

ghost:
Comfort: 4
Completeness: 4

Comments: This lab was pretty much all of the highs and lows of the week compressed into one day... I was trying for a top-down approach and realized pretty soon in that I was in over my head but slogged through the morning and into the afternoon. With James and Nayana's help, though, I was eventually able to create a viable MVP with a single event listener handling all four transaction types - which is pretty cool and definitely not something I thought I could do.

The 4s are because some of the specific functionalities - fixing to 2 decimal places, disallowing negative balances, red background at zero, etc - aren't in yet (and I'm planning to return to this one and polish it up) BUT I'm still feeling good about taking an unconventional approach and making it work

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,38 @@
+//   make sure document is ready
+//   make sure inputs are ready
+        // checking: input, deposit, withdraw
+        // savings: input, deposit, withdraw
+//   get/display checking balance
+//   get/display savings balance
+//
+// for the checking account:
+//   make sure input field is clear and ready for user input
+//   get and store value from input field
+//     make sure input is float w/ two decimals
+//   add input to balance and set if user clicks "deposit"
+//   subtract and set if user clicks "withdraw"
+//     if withdrawal request is greater than available balance, don't withdraw
+//   if the account balance is now 0, change the background to red
+//   after the transaction is complete, clear the input field, display the new balance and await user input
+
+// for the savings account:
+//   make sure input field is clear and ready for user input
+//   get and store value from input field
+//     make sure input is float w/ two decimals
+//   add input to balance and set if user clicks "deposit"
+//   subtract and set if user clicks "withdraw"
+//     if withdrawal request is greater than available balance, don't withdraw
+//   if the account balance is now 0, change the background to red
+//   after the transaction is complete, clear the input field, display the new balance and await user input
+
+// BONUS:
+// if checking account balance is less than withdrawal request...
+//   get sum of both account balances
+//   if the sum of both account balances is greater than withdrawal request...
+//     withdraw all money from checking and (withdrawal - checking balance) from savings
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,80 @@
...
+    var holdThis = $(this);
+    accountType = holdThis.parents().eq(0).attr("id");
+    if (accountType == 'checking') {
+      requestAmount = Math.round(parseFloat($('#checking .input').val()) * 100) / 100;
+    }
+    if (accountType == 'savings') {
+      requestAmount = Math.round(parseFloat($('#savings .input').val()) * 100) / 100;
+    }
+    checkTransactionType(this);
+  }
```
andrewsunglaekim, line 56: wow, this is great! 👍  Nice work on 1 event listener.

atm.js
```diff
@@ -1,3 +1,80 @@
...
+
+  function checkTransactionType(button) {
+    if ($(button).attr('value') == 'Deposit') {
+      requestAmount *= 1;
+    }
+     if ($(button).attr('value') == 'Withdraw' ) {
+      requestAmount *= -1;
+    }
+    clearTransaction();
+  }
```
andrewsunglaekim, line 66: oh this is a clever way to do deposits vs withdrawal

andrewsunglaekim:
I really like your thought processing Taylor! The code you have is really great. Really dry and modular. I was curious, though, did you start writing this code from the start or is after some refactoring? The reason I ask, is sometimes developers try and go for the really neat and maintainable solution right from the get go and falter because they couldn't find that magic solution. You did AWESOME on this assignment. However, sometimes it's totally fine to write really WET code at first just to get something working and stifle initial frustration when prototyping/getting started. Anywho, thanks as always for your awesome commentary in your journey of this particular assignment, lol. Great work! 👍

ghost:
Thanks Andy! It's really cool that you guys look at these assignments and leave such detailed feedback - I appreciate your comments here, as always.

For these first two weeks, my normal approach has been more bottom-up, organic, "well I know I can do [small piece] so let me start on that and oh hey I can use that for [other small piece] and let me see how I can connect these once I get going". For this particular lab, though, I wanted to give myself the challenge of working through a top-down approach - especially since part of the lab was to produce / incorporate pseudocode. So no, not a whole lot of refactoring; I pretty much just dove in from the top down and swam as far down as I could.

Thinking on it over the last few days, and since we're going into a project week, I think one of my biggest takeaways from this lab was seeing for myself the value of balancing between the top-down approach of pseudocode, user stories, etc against the bottom-up approach of just jumping in and getting code that works. If you've got any thoughts on how to find the midpoint between those two - or any methods / devices you find useful in your own process - I'd love to hear them :)

PR #129, opened 2016-07-07
Sarah Shultis atm

skshultis:
comfort 3/5
completeness 5/5

comments: pulled it together before 11 PM wow


automatedteller.js
```diff
@@ -0,0 +1,83 @@
...
+  var userWithdraw = $("#checking .input").val();
+  var balanceAmount = $("#checking .balance").html();
+  var nUserWithdraw = parseFloat(userWithdraw);
+  var nBalanceAmount = parseFloat(balanceAmount.split("$")[1]);
+
+  console.log(nUserWithdraw);
+  console.log(nBalanceAmount);
+
+  $("#checking .balance").html("$" + (nBalanceAmount - nUserWithdraw).toFixed(2));
+
```
nayana487, line 49: I would have included a conditional statement here to make sure this only happens if (nBalanceAmount-nUserWithdraw is going to greater than or equal to zero. Maybe then another nested conditional to add the zero class to turn the background red when it does == 0 and then don't allow anything to happen if it goes negative. Maybe send out an alert saying.."whoops you don't have the money?"

automatedteller.js
```diff
@@ -0,0 +1,83 @@
...
+//
+// });
+
+$("#checking .deposit").on("click", function(e)  {
+    e.preventDefault();
+    var userDeposit = $("#checking .input").val();
+    var balanceAmount = $("#checking .balance").html();
+    var nUserDeposit = parseFloat(userDeposit);
+    var nBalanceAmount = parseFloat(balanceAmount.split("$")[1]);
+
```
nayana487, line 29: nice work getting rid of that $

automatedteller.js
```diff
@@ -0,0 +1,83 @@
...
+  console.log(nUserWithdraw);
+  console.log(nBalanceAmount);
+
+  $("#savings .balance").html("$" + (nBalanceAmount - nUserWithdraw).toFixed(2));
+
+});
+
+
+});
```
nayana487, line 83: In general, I'd say you should be adding way more pseudocode. Thinking through every step will help you a lot when you go to write the actual code.

PR #130, opened 2016-07-07
finished ATM

EvanSanderson:
completeness: 4
comfort level: 4

Ok I know this is a huge amount of code, but I felt great being able to solve it my way (albeit in a long manner) and get the first bonus down. WOO!

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,27 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+  console.log("JS is working");
+
+  // var depositChecking;
+  // var withdrawChecking;
+  // $("#checking .input").on("change")
+  // depositChecking = $(".input").val();
+  // console.log(depositChecking);
+  // console.log(typeof(depositChecking));
+
+  $("#checking .deposit").on("click", function(e){
+      console.log("button working");
+      e.preventDefault();
+      var userAmount = $("#checking .input").val();
+      var userAmountN = parseFloat(userAmount);
+      var balance = $("#checking .balance").html();
+      var nBalance = parseFloat(balance.split("$")[1])
+      var totalBalance = nBalance + userAmountN;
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,185 @@
...
+    checkingNewTotal = $("#checking .balance").html("$" + newCheckingTotal.toFixed(2));
+  }
+
+function addToSavings() {
+  var savingsTotal = $("#savings .balance").html();
+  var savingsTotalNumber = parseFloat(savingsTotal.split("$")[1]);
+  var newSavingsTotal = savingsTotalNumber - overDrawAmount;
+  savingsNewTotal = $("#savings .balance").html("$" + newSavingsTotal.toFixed(2));
+}
+
```
nayana487, line 184: Make sure to account for the scenario where you overdraft from savings to try and withdraw from checking! Right now, it's going negative

atm.js
```diff
@@ -1,3 +1,185 @@
...
+var balance;
+var totalBalance;
+var userAmount;
+var nBalance;
+var userAmountN;
+var overDrawAmount;
+var checkingNewTotal;
+var savingsNewTotal;
+
```
nayana487, line 11: I would really avoid having this many global variables...it could mess up your code later when you don't know where scope is being applied.

atm.js
```diff
@@ -1,3 +1,185 @@
...
+    makeSavingsRed();
+    ifSWithdrawAtStart();
+    overDrawSavings();
+
+
+    $("#savings .input").val("");
+  })
+
+// savings functions
+
```
nayana487, line 92: You have a lot of functions here. I think you could tighten up your code a lot if you were to combine some of these. For example, functions that capture all the logic behind checkingDeposit, checkingWithdraw, savingsDeposit and savingsWithdraw.

nayana487:
Nice job Evan! Not a huge deal at this stage to have too much code as long as it's working.

PR #131, opened 2016-07-07
Daniela Noir HW - ATM

danielanoir:
comfort: 3
completeness: 5

comments: I finally felt like I understand what I was doing and what was working in js. I still don't feel great about it all but a lot better.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,28 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+
+$("body").css("background-color", "red");
+
+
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,28 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+
+$("body").css("background-color", "red");
+
```
superbuggy, line 4: hey i noticed you marked the assignment's completeness as a 5 but just have a couple lines of code here; do you have code locally that you need to push up?

PR #133, opened 2016-07-08
ATM Lab Courtney seitz solution

Courtney-Seitz:
Comfort-level: 3
Completeness: 4


atm.js
```diff
@@ -1,3 +1,111 @@
...
+    //$("#checking .balance").html(checkingBalance);
+    //console.log(checkingBalance);
+    $("#checking .input").val("");
+    if (balanceTest2 < 0) {
+      alert ("whoa");
+    }
+    else {
+      $("account .balance").html(("$") + (balanceTest2).toFixed(2))
+    }
+  }
```
nayana487, line 61: For both checking and saving, what you need to do is capture the val() from checking or savings input (the class) and subtract from balance using the html() method. Store this to a variable--maybe called total-- and then use this to set the new inner html of balance.

atm.js
```diff
@@ -1,3 +1,111 @@
...
+    if (balanceTest2 > 0) {
+      $("#checking .balance").removeClass("zero");
+    }
+  }
+  //click listener for function to add user input>
+  $("#checking .withdraw").on("click", updatedWithdrawal);
+  //function for checking withdrawal
+  function updatedWithdrawal (){
+    //gets user numeric input from checking account deposit>
+    var withdrawChecking = Number.parseFloat($("#checking input").val());
```
nayana487, line 42: think you meant to use the input class here in the selector.

PR #137, opened 2016-07-08
Nelley-ATM homeworkSolution

mrnelley:
Comfort: 4
Completeness: 4

Comments: This was definitely challenging... evidenced by the fact that it is still not perfect. However, I think this lab that forced me to think and just trust my code, really created a breakthrough. I am finally feeling like I can do this; like I know enough to do SOMETHING. lol

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -119,3 +119,17 @@ $(document).ready(function(){
     }
   });
 });
+
+//*******************CORRECTIONS THAT NEED TO BE MADE*******************//
+//1. if then statement for the change in color needs to be fixed on positive accounts
+//2. place entire work flow into 1 function to call on variables in both checking and savings
+//3. use named variables in both functions to refactor code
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,135 @@
...
+//3. use named variables in both functions to refactor code
+//4. create overfraft function() if then
+//5. if withdrawFromChecking.UserInput > current balance.checking:
+//6. then {
+//          Alert ("How much would you like to come out of checking?");
+//          var takefromchecking = UserInputNumber;
+//          var difference = UserInput - UserInputNumber;
+//          savings = savings.currentTotal - difference;
+//          checking = checking.currentTotal - UserInputNumber
+//        }
```
superbuggy, line 135: Looks good; only thing missing is to reject withdrawal requests that put the balance into the negative, but 👍

PR #138, opened 2016-07-08
ATM Gio

GioSilva:


atm.js
```diff
@@ -1,3 +1,118 @@
...
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+
+     console.log("Hola hola hola");
+
+// CHECKING DEPOSIT
+
+$("#checking > .deposit").on("click", function(evt){
+    // Prevent default action
+    evt.preventDefault();
+
```
nayana487, line 10: You actually don't even need the > for the selector here because there's only one deposit class inside checking. I'd only use the > id if you were worried about capturing other descendants that weren't direct children, but you don't have that here.

atm.js
```diff
@@ -1,3 +1,118 @@
...
+    var newInput = $("#checking > .input").val()
+    var floatnewInput = parseFloat(newInput).toFixed(2);
+    console.log(floatnewInput);
+
+    // update total
+      //  get current total
+    var total = $("#checking > .balance").html()
+    var numerictotal = parseFloat(total.split("$")[1])
+    var actualtotal = numerictotal - parseFloat(floatnewInput)
+    $("#checking > .balance").html("$ " + actualtotal.toFixed(2));
```
nayana487, line 47: So this code allows you balance to go negative. I would have included a conditional for this to be executed only if the resulting total would be greater than or equal to 0.

atm.js
```diff
@@ -1,3 +1,118 @@
...
+    console.log()
+    $("#checking > .input").val("")
+
+    // Turn checking box to red if balance reaches 0
+    if (actualtotal == 0) {
+      $("#checking").removeClass("account").addClass("zero")
+    }
+    })
+
```
nayana487, line 30: If the balance is going to negatives, might have kept it at red for less than 0.

PR #141, opened 2016-07-08
Rasheeda A. HW

RasheedaAlexander:
Completeness: 3
Comfort: 3

I had every intention on completing this assignment but I crashed as soon as I got home. My withdraw button turned the div red no matter what the amount was. I will work on it over the weekend.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,20 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+//User input a dollar amount to deposit money to checking
+$('body').css("background", "red");
+  //update amount
+//input a dollar amound to withdraw money for checking
+  //update amount
+//if User withdraw amount > account balance
+  //reject transaction
+//color of bank account green for positive value
+  //red for 0 (or less for bonus)
+
+  //User input a dollar amount to deposit money to savings
+    //update amount
+  //input a dollar amound to withdraw money to savings
+    //update amount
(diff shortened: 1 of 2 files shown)


RasheedaAlexander:
Also left the jQuery cdn file out. :(

atm.js
```diff
@@ -1,3 +1,74 @@
+//Checking variable
```
andrewsunglaekim, line 1: I like this getting all the variables you'll need at the top 👍

atm.js
```diff
@@ -1,3 +1,74 @@
...
+    //update amount
+    checkingInputVal = parseFloat(checkingInput.val());
+    console.log("Checking Input Value is a type of: " + typeof checkingInputVal);
+
+    if(checkingBalance > 0) {
+      $('#checking .balance').html("$" + (currentChecking - checkingInputVal));
+    }
+    else {
+      $('.balance').css("background", "red");
+    }
```
andrewsunglaekim, line 52: I think I know why it's always red:

what is the value of checkingBalance on line 47? Is it ever an integer? You defined it on line 3 but never redefined it. Its always a jQuery object and therefore isn't a number and therefore will never be greater than 0. Instead of `checkingBalance` maybe use `currentChecking - checkingInputVal` better yet, store that value in another variable, like `updatedTotal` or something.

andrewsunglaekim:
👍  Good work, definitely and try to tackle this problem again at some point when you have the time!

PR #154, opened 2016-09-16
atm Lab 02 (Mike Rubin)

merubin:
comfort level 5
completeness 5

Completed bonus - was not able to use toggle class but I set the background colors in my function setBkgActBox(account) line 40 ( I could use a hint how I might better do this and change class to use .zero class.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,40 +1,43 @@
 
 
+
+/* Global Variables */
+var actBal = {
+  "checking":200,
+  "savings":200,
+  "odProtect":false
+}
+
 /* Constant Definitions */
 const
-CHK = "checking";
-SAV = "savings";
 POSCLR ="#6C9A74";
 ZROCLR ="#F52F4F";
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,3 +1,144 @@
...
+function getActBal (account) {  /* return the current account balace of indicated account */
+  return(actBal[account]);
+}
+
+function setBkgActBox(account) { /* set account box to correct color for zero balance */
+
+   var clr; /* color to set background to */
+
+    getActBal(account)>0?clr=POSCLR:clr=ZROCLR;
+    $("#"+ account +".account").css("background-color",clr);
```
jshawl, line 45:
``` js
if(getActBal(account) > 0){
  $("#"+ account +".account").removeClass("zero")
} else {
 $("#"+ account +".account").addClass("zero")
}

```

jshawl:
👍

PR #156, opened 2016-09-16
so far

brophy2321:
completeness-1/5
comfort-2/5
had a rough week with javascript. went home last night and reread the prework which definitely made more sense.  didnt get it to actually work. was working with a friend from Cali which i think may have made things worse and more confusing. determined to finish to full completeness on my own time.


amaseda:
:+1:

PR #158, opened 2016-09-16
atm chris jaffier

chrisjaff:
completeness -3
comfort - 3

The checking side works, but the savings side doesn't yet.


amaseda:
:+1:

PR #171, opened 2016-10-12
Rob Z ATM Work

robertzehnder:
Completeness: 5
Comfort: 4

Hey @nayana487, in the withdraw function, there is logic to identify which position in the array to identify, so that I can choose an account balance based on what's passed in. Is there a way to do this without the if statement at the top? Thanks!


robertzehnder:
Same comments as before

nayana487:
I think the way you have it is a smart way to do it--another option would've been creating separate functions for each account

PR #173, opened 2016-10-13
ATM LAB

nguyendthinh:
Comfort: 4
Completeness: 5

Plus Bonus

I need help coming up with a simpler way to condense my code so I don't repeat the same iteration of code over and over again.


superbuggy:
👍 one refactoring opportunity might to condense the deposits into one function which would also take an argument that would determine which account was being accessed. from there you could even refactor further by passing yet another argument that would determine whether you were withdrawing or depositing by multiplying the number argument of the function by -1.

PR #175, opened 2016-10-13
Bradley Cravens Lab

bradleycravens:
Completeness: 5
Comfort: 4

Is there an easier way to do the overdraft protection?


andrewsunglaekim:
There are always alternatives in code. You way is totally fine for that logic. There are certainly opportunities to refactor. Set up a one - on -one with an instructor! But this code looks perfectly fine for this stage of the class. 👍

PR #177, opened 2016-10-13
Drew ATM lab

drewswinburne:
Comfort: 5
Completeness: 5

I know this could be way more "DRY", but doing it like this to get comfortable with the logic. Added overdraft protection (and overdraft fees)


nayana487:
👍

PR #178, opened 2016-10-13
Sam Catherman ATM lab

SamCatherman:
Completeness: 4.5
Comfort: 4
Comments: I spent a lot of time playing around with my overdraft function today, and I think I made it a lot worse. Submitting my last commit from yesterday when it was at least working.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,9 +1,15 @@
 $(document).ready(function(){
 
 var checkingBalance = 0;
-
+//when clicked
   $("#checking .deposit").on("click", function(){
-    checkingBalance = $("#checking .input").val();
+//store user input to variable entry
+  var entry = Number($("#checking .input").val()); //listen for user input
+    checkingBalance += entry;
+    //display entry to the checking account div
     $("#checking .balance").html("$" + checkingBalance);
+    //updates value of checkingBalance
+    $("#checking .balance").val(checkingBalance);
+    console.log(checkingBalance);
   })
 });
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,61 @@
...
+      $("#checking .balance").html("$" + checkingBalance);
+      $("#checking .balance").val(checkingBalance);
+
+      $("#savings .balance").html("$" + savingsBalance);
+      $("#savings .balance").val(savingsBalance);
+      return;
+  } else if ((checkingBalance + savingsBalance) < withdrawal){
+     alert("insufficient funds to make this transaction. Please try a different amount");
+  }
+  }
```
andrewsunglaekim, line 60: i don't think you neccessarily need `.val()` and `.html()`. `.val` is generally used for getting values out of `input` tags.

atm.js
```diff
@@ -1,3 +1,61 @@
...
+  //overdraft protection
+  function overDraft(){
+    var withdrawal = Number($("#checking .input").val());
+  if (checkingBalance < withdrawal){
+      savingsBalance += checkingBalance;
+      checkingBalance = 0;
+      console.log(savingsBalance);
+      $("#checking .balance").html("$" + checkingBalance);
+      $("#checking .balance").val(checkingBalance);
+
```
andrewsunglaekim, line 53: this is clever adding the two balances together, and you are really close, i think you just need to subtract the withdrawal amount to update the savings balance

andrewsunglaekim:
👍 Looks good!

PR #181, opened 2016-10-13
atm

pootsmagoots:
So I have a lot of random code that is written. I was practicing different ways to input the code.

comfort: 2
completeness:3

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,115 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
-});
+
+var checkDeposit = $("#checking .deposit");
+var checkWithdraw = $("#checking .withdraw");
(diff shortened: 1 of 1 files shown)


atm.js
```diff
@@ -1,3 +1,115 @@
...
+var inputChecking = $("#checking .input");
+var inputSaving = $("#savings .input");
+var storedChecking = $("#checking .balance");
+var storedSaving = $("#savings .balance");
+
+checkDeposit.on("click", function(evt){
+ evt.preventDefault();
+ var amount = $("#checking .input").val();
+ $("#checking .balance").html("$" + amount);
+})
```
andrewsunglaekim, line 16: i think you're really close, really just have to add the amount to what already exists in the balance.

andrewsunglaekim:
👍 Come see an instructor during one - on -one and we can work through this problem. I think you're really close!

PR #183, opened 2016-10-13
Sam Mans ATM HW

sdmans:
Comfort: 3
Complete: 3

Comment: I couldn't get the decimals to display and carry over or get the colors to change based on the value


superbuggy:
try using parseFloat instead of parseInt. 👍

PR #184, opened 2016-10-13
ATM Lab(Ba Phan)

BaVPhan:
Completeness: 3
Comfort: 3

Tried to get color changing and overdraft functionality, only got some features to work, not everything. Overdraft concept got me a bit frustrated cause some parts would break my code. gonna cry to sleep tonight


superbuggy:
it's tricky at first to think through the program's execution step by step. but it seems like maybe it wouldn't be as hard for you the next time around! 👍

PR #185, opened 2016-10-13
Alex Wandres ATM

awandres:
Comfort - 4
Completeness - 5

everything functions as expected/requested. Had a few hangups but overall felt pretty confident that I knew what my script was doing and why


nayana487:
👍

PR #189, opened 2016-10-14
Diana's ATM

dbgobar:
completeness: 4.5 
comfort: 4

Something is wrong with my savings withrawl portion of the code..but other that it works


andrewsunglaekim:
👍  not really sure what's causing that error, seems like it would be consistent with the rest of your app. I'm happy to debug this with you though if you'd like. Set up a 1 - on - 1 if you want to figure it out!

PR #191, opened 2017-01-27
Nat Carter | ATM

ghost:
Comfort: 5 / 5
Completeness: 5 / 5

Any feedback on my refactoring approach or my comments/pseudocode (e.g., am I over-commenting? are my comments too similar to the actual code to be helpful?) would be appreciated, but I feel pretty good about this.


amaseda:
:+1: Great job. As to your comments, I think for this assignment the way you included your pseudocode is great. 

In the future, rather than give every single line a comment I think it will make your code much more readable if you just precede each relevant function or chunk of code with some words on what it does. You can leave the developer to dive into it line by line (although if there's a particularly dense line, you can leave additional comments).

PR #198, opened 2017-01-27
Clements ATM-Lab

sdkclements:
Got an MVP working
creating pull request early

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -51,6 +51,9 @@ $(document).ready(function() {
 
 
 });
+// function takefromOther(){
+	// if this.withdrawValue > 
+//}
 
 
 function zeroCheck() {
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,142 @@
...
+
+function withdrawSavings() {
+    ballance = parseInt($('#savebal').text())
+    valtoadd = parseInt($('#inputSaving').val())
+    if (valtoadd > ballance)
+        alert('you cannot take out that much');
+    else
+        updated = (ballance - valtoadd);
+    $('#savebal').text(updated);
+}
```
amaseda, line 93: As you probably noticed, the above two functions look pretty similar. How could you consolidate them into a single function? Perhaps you could make the balance and value arguments that vary depending on whether it's a checking or savings account...

amaseda:
:+1:

PR #208, opened 2017-01-27
Aaron Andrews

andrews94:


atm.js
```diff
@@ -1,3 +1,91 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+
 });
+
+checkingDeposit = $("#checking").find('.deposit');
+checkingWithdraw = $("#checking").find('.withdraw');
+savingsDeposit = $("#savings").find('.deposit');
+savingsWithdraw = $('#savings').find('.withdraw');
+checkingBalance = 0;
```
superbuggy, line 9: I'd move the above 4 lines inside the document-ready anonymous call-back block. Sometimes, the page won't have finished loading and $ attempts to select an element that hasn't been rendered to the page yet. 

You have your script tag at the bottom, which mostly ensures this won't happen, but doc.ready callbacks are more reliable.

Also use keyword `var` for declaring variables to avoid the use of declaring globals.

atm.js
```diff
@@ -1,3 +1,91 @@
...
+function redChecking () {
+if (checkingBalance === 0) {
+  $('#checking .balance').addClass('zero');
+}
+else {
+  $('#checking .balance').removeClass('zero');
+  }
+}
+redChecking();
+
```
superbuggy, line 75: You can also refactor this to take the account type as an argument, if you want to boil these functions above and below into 1 functions

atm.js
```diff
@@ -1,3 +1,91 @@
...
+// update savings balance with deposit
+function updateSavingsDeposit () {
+  savingsBalance += parseInt($('#savings .input').val());
+  $("#savings .balance").html('$' + savingsBalance);
+  $('#savings .input').val("");
+  redSavings();
+}
+
+
```
superbuggy, line 49: I like that you've broken the updating out into functions. Very common pattern in development.

atm.js
```diff
@@ -1,3 +1,91 @@
...
+function redSavings () {
+if (savingsBalance === 0) {
+  $('#savings .balance').addClass('zero');
+}
+else {
+  $('#savings .balance').removeClass('zero');
+  }
+}
+redSavings();
+
```
superbuggy, line 86: Comments are always nice! I'd move the event listener adding block below to the doc.ready block up top.

atm.js
superbuggy, line 9: really the find methods are unnecessary $(".checking .deposit") is enough; again, the string passed into $ is just a css selector. Using `.find()` is an unnecessary computational expense.

superbuggy:
👍 overall

PR #209, opened 2017-01-27
ATM Lab - Erica Norby

ericanorby:
Comfort: 4
Completeness: 5

Fun exercise. Great practice using JS. I feel like I could have refactored the code even more but not sure how to do that.


phantomhaircuts:
👍

PR #215, opened 2017-01-27
Bedig Galladian ATM

bgalladian:
Comfort 3
Still need a lot of work in javascript, but feel a little better.  My group definitely helped.

Completeness 4
I'm sure I'm missing something...but my brain is fried.


AndyWhitley:
👍

PR #217, opened 2017-01-27
ATM Savraj Singh

SavrajSingh:
Completedness:4
Comfort:2 initially 3.5 at latest commit

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -16,11 +16,12 @@ var originalBalance = 0
 
 var depositChecking = $("#checking .deposit");
 depositChecking.on("click", function(){
-      var prevBalance = $("#checking .balance").val();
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,89 @@
...
+
+        $("#checking .balance").html("$" + newBalance)
+
+    })
+
+
+var withdrawChecking = $("#checking .withdraw")
+
+withdrawChecking.on("click", function(){
+  var prevBalance = $("#checking .balance").val().replace("$","");
```
amaseda, line 34: This code looks very similar to what you have in your deposit checking listener. Perhaps this is an opportunity for refactoring. Think about how you could use `this` to do that...

amaseda:
:+1:

PR #222, opened 2017-01-27
Clark ATM

nrc0004:
img_2782
img_2783 2

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -18,8 +18,23 @@ $(document).ready(function(){
     var withdrawBalance = parseInt (balance) - parseInt( $("#depo").val())
 
     $(".checkingbalance").html(withdrawBalance)
+  });
 
+  $(".savingdeposit").click(function(){
 
+    var savingbalance = $(".savingbalance").html()
+    var depositBalanceSaving = parseInt (savingbalance) + parseInt( $("#savingdepo").val())
+
+    $(".savingbalance").html(depositBalanceSaving )
 
   });
-});
+
+  $(".savingwithdraw").click(function(){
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,3 +1,64 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
-});
+
+  var value =0
```
amaseda, line 3: Do you use this anywhere?

atm.js
```diff
@@ -1,3 +1,64 @@
...
+  var balance
+  // var depositBalance = parseInt (balance) + parseInt( $("#depo").val())
+  // var savingbalance = $(".savingbalance").html()
+
+  $(".deposit").click(function(){
+
```
amaseda, line 9: You might want to save the value of `$("#depo").val()` to a variable inside of this event listener, since you already do it with `balance`. Would make your code more uniform.

atm.js
```diff
@@ -1,3 +1,64 @@
...
+    var balance = $(".checkingbalance").html()
+    var depositBalance = parseInt (balance) + parseInt( $("#depo").val())
+
+    $(".checkingbalance").html(depositBalance)
+    if (balance + ($("#depo").val()) >= 0) {
+      $("#checking.account").css("background", "rgb(108,154,116)");
+    } else {""}
```
amaseda, line 16: You don't need to include this `else` statement if it doesn't do anything. Just an `if` is fine.

amaseda:
:+1:

PR #226, opened 2017-01-27
Nyketha's ATM LAB

Nykke:
level of completeness: 4 ... ish? 

level of understanding: 3

comments: I tried, this was very difficult for me, jQuery.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -11,34 +11,32 @@ $(document).ready(function(){
   //Users of the ATM object can deposit money into bank account
   // Set the value of 0
 
-  var checkingAccount = 0;
-  var checkingBalance = ('checking', div.balance);
-  var checkingInput = ('input', input.input);
-  var checkingDeposit = ('button', input.deposit);
-  var checkingWithdraw = ('button', input.withdraw);
+  // var checkingAccount = 0;
+  // var checkingBalance = ("checking", div.balance);
+  // var checkingInput = ("input", input.input);
+  // var checkingDeposit = ("button", input.deposit);
+  // var checkingWithdraw = ("button", input.withdraw);
 
   //Users of the ATM object with a Savings Account
   //Users of the ATM object can deposit money into bank account
   //Set the value of 0
 
-  var savingsAccount = 0;
-  var savingsBalance = ('saving', div.balance);
-  var savingsInput = ('input', input.input);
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,122 @@
...
+
+//When withdrawn button is clicked it will return the userinput
+//When userinput is returned it will console.log('ballin')
+//When the withdraw button is clicked, the balance container will update (in add. to current) with userinput
+
+$("div#savings > .deposit").on("click", sumTotal);
+function sumSavingsTotal() {
+  var checkingBalance = parseInt($("#savings > .input").val());
+    if($(this).val()==$(this).attr('.deposit'))
+         $(this).val().split();
```
superbuggy, line 101: I'm not sure what you're trying to do here, but $(this).attr('class') is what would give the account type in this example. there isn't an html attribute called `.deposit`.

`.split("")` creates an array out of a string. I think you were trying to determine whether or not you'd be able to withdraw from your account based on the balance. 

Guessing just plain doesn't work in programming. You need to be very specific and intentional about what you are doing. It's part of the difficulty of the work!

This is one reason why it makes sense to pseudo code, but also declaring variables with meaningful names like `amountRequested` and `checkingBalance`. Pseudocode or identify the information you need, devise names for your variables that reflect what information they hold, and then define them with the appropriate data.

Easier said than done, but what I want to stress is that we, as programmers, are forced to repeatedly break down problems with accuracy and specificity. 

I'd like you to pseudocode when you have a larger assignment or lab, because I think it would help you simplify/clarify your thinking about a certain problem.

atm.js
```diff
@@ -1,3 +1,122 @@
...
+//When userinput is returned it will console.log('ballin')
+//When the withdraw button is clicked, the balance container will update (in add. to current) with userinput
+
+
+("#checking > .withdraw").on("click", sumTotal);
+
+function sumLoss() {
+  var checkingBalance = parseInt($("#checking > .input").val());
+    if($(this).val()==$(this).attr('.withdraw'))
+         $(this).val().split();
```
superbuggy, line 85: `.val()` will only grab the text typed into inputs where as `.html()` will grab the inner HTML of a DOM element.

superbuggy:
👍 effort overall!

PR #230, opened 2017-01-27
RODNEY CURL — Submission ATM (LAB I)

rodneycurl:
completeness: 4/5
comfort level: 5

I got as far as I could with what I understood in the short amount of time. I'm going to work on this some more tonight just to see it through to completion but def wanted to send something over.

—Rodney


nayana487:
👍

PR #249, opened 2017-05-10
progress thus far

BaoPham92:
I have used a parseInt in the js, how do I actually deal with the "$" being removed? I also realized I should have created a issue instead of a PR. I attempted to remove the PR, however I may not be as familiar with Github functionality perhaps, but I don't know if I cancel it haha.

### DIFF
--- a/atm.css
+++ b/atm.css
@@ -6,6 +6,7 @@ body {
   margin: 0px auto;
   text-align: center;
   width: 700px;
+  background: red;
 }
 
 .header {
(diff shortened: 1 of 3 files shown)


BaoPham92:
So I forgot, but on the latest commit for the file. I was curious if you could look at the failSafe variables I have created at the bottom of the script file. I tried other methods but so far that was the first one one that actually prevented the balance from going negative. The only problem though is that the the value stops at '0' for depositing as well.

nayana487:
Look into the `replace` method to figure out how to get rid of the $ sign

nayana487:
Why are you storing a conditional in a variable? Do you want that to be a function? I see you're declaring and setting the value of the variable, but don't see how it's actually being used. So nothing is happening with that at the moment.

PR #250, opened 2017-05-12
created Checking and Saving functions

Fitssum:


index.html
```diff
@@ -14,18 +14,19 @@
...
     <div id="savings" class="account">
       <h2>Savings</h2>
-      <div class="balance">$0</div>
-      <input class="input" type="text" placeholder="enter an amount" />
-      <input class="deposit" type="button" value="Deposit" />
-      <input class="withdraw" type="button" value="Withdraw" />
+      <div id="savingBalance">$0</div>
+      <input id="savingAmount" type="text" placeholder="enter an amount" />
+      <input id="savingDepositt" type="button" value="Deposit" />
+      <input id="withdrawDeposit" type="button" value="Withdraw" />
```
superbuggy, line 28: For this assignment, you had to just write javascript but not change the html. There will often be times as a developer that you cannot change the html that your javascript is interfacing with.

The key here is how to select elements (write selectors) and use them to select elements in jQuery like: `$(".balance .input")`

Fitssum: Got it thanks for the feedbacks

superbuggy:
Everything else looks good! 👍

PR #253, opened 2017-05-12
atm lab

usmanahmed090:


atm.js
```diff
@@ -1,3 +1,150 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
-});
+ //DEPOSIT MONEY FROM CHECKING
+
+
+
+  var checkingDeposit = $("#checking .deposit")
+  var savingsDeposit = $(“#savings .deposit”)
```
amaseda, line 7: Please make sure you're using the correct quotation marks in your code. I notice the ones on this line are a little different than the previous line.

Not sure why this happens sometimes but happy to look into it for you if it continues to be an issue.

atm.js
```diff
@@ -1,3 +1,150 @@
...
+
+
+  checkingDeposit.on("click", function(){
+
+
+    var checkingUserInput = $("#checking .input").val()
+    console.log("This is the user input:" + checkingUserInput)
+
+     create a variable that stores "checking"‘s balance(in this exercise = $0)
```
amaseda, line 16: This looks like it's supposed to be a comment but you didn't precede it with `//`. This will cause your code to break. Were you able to properly test your code?

atm.js
```diff
@@ -1,3 +1,150 @@
...
+    var checkingOutputBalance = $("#checking .balance")
+
+
+    checkingBalance += parseInt(checkingUserInput)
+    console.log("This is the checking balance " + checkingBalance)
+
+     display "$" + "checkingBalance" ($0, or whatever user inputs in our balance ("outputBalance"))
```
amaseda, line 23: This looks like it's supposed to be a comment as well.

atm.js
```diff
@@ -1,3 +1,150 @@
...
+ console.log("This is the checking balance " + checkingBalance);
+
+ checkingOutputBalance.text("$" + checkingBalance);
+
+
+
+
+ var savingsDeposit = $("#savings .deposit")
+
+ savingsDeposit.on("click", function() {
```
amaseda, line 54: Please indent your code properly so it's easy to tell, for example, what code is inside of this click listener's callback function.

amaseda:
There are some lines in your code that will prevent everything from properly running. Please look through my comments and let me know if you have any questions/comments.

@usmanahmed090

PR #254, opened 2017-05-12
Tim Chase atm Solution

Tchase44:


atm.js
```diff
@@ -1,3 +1,106 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
+	var userAccounts = {
```
amaseda, line 2: Nice! What are other ways you could incorporate data/functionality into an object?

atm.js
```diff
@@ -1,3 +1,106 @@
...
+
+		moneyCheck()
+		inTheRedCheckings()
+	})
+	//deposit saving
+	$('#savings .deposit').on("click",function(){
+
+		moneySave()
+		inTheRedSavings()
+	})
```
amaseda, line 18: Any thoughts as to how you could go about creating a single event listener for deposit functionality, regardless of how many accounts there are?

Maybe you could pass in an argument to `moneySave` that indicates what account you're making a deposit into...

atm.js
```diff
@@ -1,3 +1,106 @@
...
+	$('#savings .withdraw').on("click", function(){
+
+		withdrawlSave()
+		inTheRedSavings()
+	})
+
+
+//////////////////////////////////////////////////////////
+	//function to withdrawl $$$ from checkings
+	function withdrawlCheck() {
```
amaseda, line 35: I like your function-based approach!

amaseda:
:+1:

PR #257, opened 2017-05-12
final modification

tedlayea:


atm.js
```diff
@@ -1,3 +1,70 @@
 $(document).ready(function(){
-// Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
-});
+
+var type= prompt("For saving account enter 1 or for saving account enter 2, please")
+if(type==1)
```
amaseda, line 4: This works, but remember, the `1` is coming in as a string via the `prompt`

atm.js
```diff
@@ -1,3 +1,70 @@
...
+{
+  var depositButton = $("#checking .deposit")
+  var withdrawButton =$("#checking .withdraw")
+  var balance1 =$("#checking .balance")
+  var balance =balance1.html().replace("$","")
+  var amount1 = $("#checking .input");
+}
+else if (type==2){
```
amaseda, line 12: Any thoughts on how you would implement this without having to use a `prompt` to determine which account is being interacted with?

amaseda:
:+1:

PR #260, opened 2017-05-12
not done but I am finished

AskForTheWolf:
I know the purpose of the assignment was to change CSS/HTML in js but the way it was created, to me, wasn't how i would have set it up.  Why create so many HTML objects (buttons) but not even throw the onClick events at it at the same time?  If I were making this from scratch I would at least assign them different names based on checking and savings (which I did to the HTML to reflect that) if I didnt already attach the event to them.  That way it would be easier to fetch and modify when necessary.

I left all my psuedocode in there to indicate what I was trying to do, but based on the wording and instructions and me being not good at this, I wasn't able to complete it.

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,3 +1,30 @@
 $(document).ready(function(){
 // Uh oh -- it's saying `$` is undefined! Something's missing from `index.html`...
 });
+
+var checkingAcct = $("#checking");
+var savingsAcct = $("#savings");
+var checkDeposit = checkingAcct.deposit;
+var checkBalance = checkingAcct.balance;
+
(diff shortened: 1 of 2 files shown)


superbuggy:
When you are writing DOM interactions with javascript best practice is to not use `onClick`; the idea here is to keep behavioral logic separate from display logic.  Very often, developers must write javascript to fit an existing interface that they will not be able to modify. This is often frustrating, but necessary to work around. I noticed that you didn't get to any DOM manipulation in the assignment, please let me know ASAP if this still something you are not comfortable with.

PR #261, opened 2017-07-10
Alex W ATM

awandres:
comfort - 5
complete - 5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -4,34 +4,67 @@ var cDeposit = $('#checking .deposit')
 var cWithdraw = $('#checking .withdraw')
 var cBalance = $('#checking .balance')
 var cInput= $('#checking .input')
+var cAccount = $('#checking')
 var cBalanceNum = 0
 
 var sDeposit = $('#savings .deposit')
 var sWithdraw = $('#savings .withdraw')
 var sBalance = $('#savings .balance')
 var sInput = $('#savings .input')
+var sAccount = $('#savings')
 var sBalanceNum = 0
 
 cDeposit.on('click', function() {
   cBalanceNum += parseInt(cInput.val())
   cBalance.text('$' + cBalanceNum)
+  if (cBalanceNum > 0) {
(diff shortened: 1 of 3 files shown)


atm.js
```diff
@@ -1,3 +1,75 @@
...
+
+cWithdraw.on('click', function() {
+  if (cBalanceNum >= parseInt(cInput.val())) {
+  cBalanceNum -= parseInt(cInput.val())
+  cBalance.text('$' + cBalanceNum)
+  cInput.val('')
+    if (cBalanceNum === 0) {
+      cAccount.attr('class', 'zero')
+    }
+}
```
juancgarcia, line 34: Watch your indentation here, lines 28-34 should be inset one level or it can confusingly look like the click function ends early.

atm.js
```diff
@@ -1,3 +1,75 @@
...
+  else if(cBalanceNum < parseInt(cInput.val())) {
+    if (cBalanceNum + sBalanceNum < parseInt(cInput.val())) {
+    alert('That amount exceeds your available balance in both accounts')
```
juancgarcia, line 37: Indentation reminder. `alert()` should be inset one more level to show it's inside the `if` block

```js
if (cBalanceNum + sBalanceNum < parseInt(cInput.val())) {
    alert('That amount exceeds your available balance in both accounts')
}
```

atm.js
```diff
@@ -1,3 +1,75 @@
-$(document).ready(function(){
```
juancgarcia, line 1: Ideally we'd want to keep our jQuery code inside of a `$(document).ready()` block for a couple of reasons. First we may not always have control of the html document to re-order script tags, (that could be a different team's baby) and our code would fail since the document wasn't fully loaded. But this also creates a scope for our variables and keeps them from polluting the global namespace, or worse colliding with other globals that might have been defined by another part of the app.

juancgarcia (commented):
👍 Good Job!

PR #263, opened 2017-07-10
Jessica Marie Barre - Atm Lab

jessamarie:


juancgarcia:
👍 Good Job!

Looks like you had no problems with this lab, but in the future, get into the habit of supplying the Comfort & Complete ratings with your submissions. That will be helpful for us when tougher assignments come up so we can suggest more assistance if needed.

PR #264, opened 2017-07-10
first commit of ATM lab

preese13:
difficulty 3/5
completeness 4/5

I built too much without consistent testing.  at this point, my tests have been unable to help me understand the errors holding my code back.  However, after walking through the code multiple times I am sure it is extremely close to functionality, I just cannot find the mistakes


jsm13:
👍 Looking really close to me too. Good observation about iterative development. Good use of a class here

PR #269, opened 2017-07-10
Charles Cox ATM

cwcox:
Complete- 5
Comfort - 4.5

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -1,25 +1,70 @@
-
+// Variables
 var checkWithdrawBut = $('#checking').children('.withdraw')
 var checkDepositBut = $('#checking').children('.deposit')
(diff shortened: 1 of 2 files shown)


atm.js
```diff
@@ -1,3 +1,110 @@
-$(document).ready(function(){
```
juancgarcia, line 1: Ideally we'd want to keep our jQuery code inside of a `$(document).ready()` block for a couple of reasons. First we may not always have control of the html document to re-order script tags, (that could be a different team's baby) and our code would fail since the document wasn't fully loaded. But this also creates a scope for our variables and keeps them from polluting the global namespace, or worse colliding with other globals that might have been defined by another part of the app.

juancgarcia (commented):
👍 Good Job!

PR #272, opened 2017-07-10
ATM-KA

kadkins4:
comfort: 3
completion: 3.5

started off strong with the class but fell off due to multiple functions needed


jsm13:
👍 Good work! If there are any specifics you want to review, let me know

PR #273, opened 2017-07-10
ATM Lab - Liz Jewell

lizardbird:
comfort: 4
complete: 5

I really wanted to create my functions in one master class, but I was struggling like whoa to get the jQuery elements to work within the class itself. So, abandoned that effort and submitting very duplicative code! :(

### DIFF
--- a/atm.js
+++ b/atm.js
@@ -17,7 +17,7 @@ var savingsInput = $('#savings .input')
 balanceChecking.addClass('zero')
 balanceSavings.addClass('zero')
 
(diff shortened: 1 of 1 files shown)


AndyWhitley:
What you would probably want to do is to create an `Account` or `Atm` class that you would create an instance of for **each** account (i.e. there would be one instance for the checking account and another for the savings account). That way, you could have only one set of functions for deposit, withdraw, etc. but they would be specific to their account. 

Nonetheless, good work 👍

PR #276, opened 2017-07-10
Sanjay_Persaud_added ATM logic

sanjaywallah:


jsm13:
👍 Great job! How're you feeling on this?

sanjaywallah:
I made the grave error of not reading through this problem, really thinking
about it, and then trying to solve it!  Got caught up in the solution
fever!  I will not make that mistake again.  It's so important to
understand these problems in plain vanilla english.  I need to state the
problem to myself, and then g about implementing it in code!

I got help from my boy, Andy.  Initially, my code was not very DRY at all,
not one bit!  I get caught up in this whole hoisting business for some
stupid reason.  I just need to think clearly about scope.

